Issue #41, opened 2017-12-06
wrong python path in virtualenv


tjingrant:
Hi, 

I use virtualenv and when I install netron, the python path is wrong in the netron executable script. I have to change that first line in the netron script (which indicates where python is located) to the python executable in the virtualenv. Just to let you know this could be a nuisance for others using virtualenv...

Best wishes,
Tian Jin.

tjingrant:
Better to dynamically generate this line https://github.com/lutzroeder/Netron/blob/10d2b8cb795ec70a7e686b49df6cb49800da6aa3/src/netron.py#L1.

lutzroeder:
Definitely. The installer is using `bdist_wheel` so not sure how to generate it. Looks like using `#!/usr/bin/env python` instead of `#!/usr/bin/python` would make this work?

tjingrant:
Seems like a good idea and it works for me.

Issue #39, opened 2017-12-06
Int64 toString() in JavaScript


lutzroeder:

BrandonDenton:
Taking a glance at your code, and I'm thinking this formatting can be done with a simple addition of NumPy. Would you mind if I added that in my fork? If there's a native tool for doing 64 bit ints, I've forgotten how to use it.

lutzroeder:
The issue is with the JavaScript viewer. JavaScript can not represent tensors with large Int64 values as Number. Currently reading them as byte[] into in Int64 in view.js. The toString() for this class needs to be implemented.

BrandonDenton:
Thanks for the clarification. Working on that and a couple other listed issues now.

lutzroeder:
The issue is that JavaScript cannot represent large 64 bit integers as they are converted to Number: https://stackoverflow.com/questions/5353388/javascript-parsing-int64. They can be loaded as byte[] but rendering them will probably involve writing custom code to render each digit by running mod 10 and div 10 on the byte array.

PR #43, opened 2017-12-07
Issue #39: Int64

BrandonDenton:
Please disregard the path fix in the earlier commit for issue #41 as you appear to have already resolved that. If the toString() isn't what you're looking for, let me know.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -402,7 +402,7 @@ ModelService.prototype.openBuffer = function(buffer, identifier, callback) {
 }
 
 function Int64(data) {
-    this.data = data;
+    this.data = String.fromCharCode(parseInt(data, 2));		// byte data converted to string
 }
 
 Int64.prototype.toString = function() {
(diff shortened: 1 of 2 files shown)


lutzroeder:
See comment added to the original issue. JavaScript cannot represent large 64 bit integers as they are converted to Number: https://stackoverflow.com/questions/5353388/javascript-parsing-int64. The issue is how to render the byte array into a string without using parseInt etc, will probably require custom code.

src/view.js
```diff
@@ -402,17 +402,17 @@ ModelService.prototype.openBuffer = function(buffer, identifier, callback) {
 }
 
 function Int64(data) {
-    this.data = data;
+    this.data = String.fromCharCode(parseInt(data, 2));		// byte data converted to string
```
lutzroeder: Since this is losing information converting from int64 byte[] to Number probably shouldn't be done in the construct. It's more of a stop gap that could be applied in toString() until real code is available that supports > 2^53 -1.

PR #69, opened 2018-01-25
First PR from walrusmcd - cosmetic , docs, local building artifacts, and update to ONNX metdata docs

walrusmcd:


setup.cfg
```diff
@@ -2,4 +2,4 @@
 universal=1
 
 [egg_info]
-egg_base = /tmp
+egg_base = tmp
```
lutzroeder: The reason this exists is so the `.egg-info` doesn't get copied into the root folder. Merge to TIP will use `./build/python` instead. Does that work?

README.md
```diff
@@ -48,6 +48,21 @@ Below are a few model files you can download and open:
...
+```
+nmake install
+```
+	
+To launch the Netron app from the current cloned sources directory
+```
+nmake start
+```
+
+
```
lutzroeder: 'README.md' is intended to only have install instructions for users. Plan is to have a different document for dev instructions at some point.

PR #74, opened 2018-02-08
Fixed auto open on win32, parsing the filename extension 

walrusmcd:


lutzroeder:
PR contains a bunch of commits that are empty. Maybe reset and start fresh?

src/app.js
```diff
@@ -52,7 +52,9 @@ class Application {
 
         if (process.platform == 'win32' && process.argv.length > 1) {
             process.argv.slice(1).forEach((arg) => {
-                if (!arg.startsWith('-') && arg.split('.').pop() != 'js') {
+                var hasExtension = (arg.split('.').length > 1);
+                var extension = arg.split('.').pop();
```
lutzroeder: What command line is causing this issue? Filter for "!startsWith('-')" first as all switches can be ignored and shouldn't be checked for extensions.

PR #97, opened 2018-03-28
support onnx opset

fumihwh:
Support onnx opset.
Show docs by version which get from model.

### DIFF
--- a/src/onnx-model.js
+++ b/src/onnx-model.js
@@ -238,7 +238,6 @@ class OnnxNode {
     constructor(graph, node) {
         this._graph = graph;
         this._node = node;
-        this._opset = this._graph._model._model.opsetImport[0].version
     }
 
     get operator() {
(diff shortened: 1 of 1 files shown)


lutzroeder:
1) What are you actually trying to accomplish with this change? Can you provide an example?
2) Instead of caching the property on each node, maybe cache it at the graph. Memory usage.
3) Do older models always have opsetImport[0] set or will this fail?

fumihwh:
1. To show documents by different version of opset.
    Currently, even I open an older model, the latest documents of operator will show up. I want to fix this problem.
    For example, reshape op change the specification that move attr shape to input. 
    When I open a latest model, documents match perfectly.
image
    But When I open an older model say opset=4, documents wrong.
image
    You can try following pbs.
[onnx_reshape_version_4](https://www.dropbox.com/s/a5mh2sq8sty767r/onnx_reshape_4.pb?dl=0)
[onnx_reshape_version_6](https://www.dropbox.com/s/80q6nqit4llz2mi/onnx_reshape_6.pb?dl=0)
2. DONE. remove caching as property and add get opset method
3. DONE. improve compatibility of older model (default opset=1)
4. I don't get it. How to do multiple imports?

lutzroeder:
Thank you. I pushed a slightly different version that includes some caching and also works if multiple `opSetImport` are defined. Please let me know if anything is missing.

lutzroeder:
[reshape_opset.zip](https://github.com/lutzroeder/netron/files/2592375/reshape_opset.zip)

PR #98, opened 2018-03-29
Fix ONNX import schema getter bug

fumihwh:
The getter has bug.
It can not get the latest version smaller than or equal to current opset.

### DIFF
--- a/src/onnx-model.js
+++ b/src/onnx-model.js
@@ -712,7 +712,7 @@ class OnnxOperatorMetadata {
             for (var i = 0; i < domainKeys.length; i++) {
                 var domain = domainKeys[i];
                 var versionMap = domainMap[domain];
-                var importVersion = imports[domain];
+                var importVersion = Math.max.apply(Math, Object.keys(versionMap).map(Number).filter(x => x <= imports[domain]));
                 schema = versionMap[importVersion];
                 if (!schema) {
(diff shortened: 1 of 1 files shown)


lutzroeder:
Do you have an example for when this happens?

fumihwh:
Operators do not have continuous number of opset versions, says Reshape op has 1 and 5.  More strictly, opset version of op is 'since_version'. We have to use the biggest number smaller than or equal to model opset.

Assume we have a model with opset version 4 contains Reshape op.
Again, Reshape op has [1, 5]. So in that case, we should use Reshape op 1.
1 is the biggest number belongs to [1, 5] <= 4.

Otherwise, if model with opset version 6, we should use Reshape op 5.
5 is the biggest number belongs to [1, 5] <= 6.

lutzroeder:
That is what the code right after in `if (!schema)` is doing. Finding the next best match and caching it -- is it not working?

fumihwh:
@lutzroeder 
Sorry, this pr should be a refactoring, not bug fix.
With my code, we don't need if branches below. What do you think?

lutzroeder:
The idea is to have cache lookups instead of iterating over lists each time which might get expensive for large models. Ideally this caching should happen at the graph level so might be tweaking this in the future.

PR #185, opened 2018-11-17
Python: Fix `netron.main()` and `netron` command without arguments

kkk669:
This PR solves the problem that `netron.main()` returns `TypeError: stat: path
should be string, bytes, os.PathLike or integer, not NoneType`. This error has been occurs since the commit ed1846e1372cc5911b6ab4a9ebe620a52911150e.

Also, you added an assert statement in [this line](
https://github.com/lutzroeder/netron/commit/ed1846e1372cc5911b6ab4a9ebe620a52911150e#diff-94ab7c76892f0032ee9ad040e3e8f207R201), but I deleted it. It won’t become `False` because `type(file)` returns NoneType not `None` if `file` is `None`, and in that case, Netron shows the file picker screen. Therefore, I suppose that it is not needed.

If I missed something, please tell me that. Thanks.

I tested on Python 2.7 and 3.6.

### DIFF
--- a/src/server.py
+++ b/src/server.py
@@ -198,7 +198,6 @@ def serve(file, data, verbose=False, browse=False, port=8080, host='localhost'):
         host (string, optional): Host to serve. Default: localhost
     '''
     global thread_list
-    assert type(file) != None
     if not data and not os.path.exists(file):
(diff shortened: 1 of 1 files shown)


lutzroeder:
Thank you!

PR #189, opened 2018-11-20
OpenVINO Toolkit Intermediate Representation Model Support

demid5111:
Hi @lutzroeder ,

## Motivation

I would like to add support for the [OpenVINO Toolkit](https://software.intel.com/en-us/openvino-toolkit) models.

The are in the special format: [Intermediate Representation (IR)](https://software.intel.com/en-us/articles/OpenVINO-ModelOptimizer#intermediate-representation-notation-catalog).

This is extremely important for our product (OpenVINO) to have a visualization engine for our [Open Model Zoo](https://github.com/opencv/open_model_zoo/blob/2018/intel_models/index.md).

So, as it is critical to have the tool live till the mid-December, I am ready to address all the issues ASAP to make sure we can have fork merged in the mainline. 

It would be a bad decision to keep this visualization as a fork and maintain it both for us and for your tool. Instead I propose to collaboratively improve it in long-term: optimizing, adding new features etc.

## How to proceed
 
Please guide me through the process - do I need to write any additional tests, go through static code analysis and so on.

## Required improvements (for future PRs)

1. Put OpenVINO XML/DOT parser generation logic in the `./tools` directory.
2. Add support for the `-metadata.json` file for better filtering of properties, group coloring etc.


electron-builder.yml
```diff
@@ -50,6 +50,12 @@ fileAssociations:
     ext: prototxt
   - name: "Caffe Model"
     ext: prototxt
+  - name: "OpenVINO IR Model"
```
lutzroeder: `xml`, `bin` and `dot` are rather general extensions reserved for other tools. They should not be registered to this app. Please remove.

lutzroeder: Do you have some context why there are 3 new formats needed. Can't OpenVINO support ONNX or one of the other existing formats instead of introducing multiple new ones?

package.json
```diff
@@ -21,6 +21,7 @@
         "electron-updater": "^4.0.0",
         "flatbuffers": "latest",
         "handlebars": "latest",
+        "lodash": "^4.17.11",
```
lutzroeder: Remove loadash dependency.

setup.py
```diff
@@ -82,12 +83,25 @@ def package_version():
         package_manifest = json.load(package_file)
         return package_manifest['version']
 
+supported_frameworks = [
```
lutzroeder: Remove.

setup.py
```diff
@@ -119,6 +133,22 @@ def package_version():
             'sklearn.js', 'sklearn-metadata.json',
             'tf.js', 'tf-metadata.json', 'tf-proto.js', 
             'tflite.js', 'tflite-metadata.json', 'tflite-schema.js', 
+
+            # OpenVINO IR
+            'openvino-ir/openvino-ir.js',
```
lutzroeder: Pattern so far is to have a single `xxxx.js` and `xxx-metadata.json` for each framework. This should be kept simple and easy to maintain. Try to fit it into a single `openvino.js` and `openvino-metadata.js` file like the other frameworks. Right now your approach seems more complex than TensorFlow or PyTorch.

src/app.js
```diff
@@ -112,7 +112,32 @@ class Application {
         var showOpenDialogOptions = { 
             properties: [ 'openFile' ], 
             filters: [
-                { name: 'All Model Files',  extensions: [ 'onnx', 'pb', 'h5', 'hdf5', 'json', 'keras', 'mlmodel', 'caffemodel', 'model', 'meta', 'tflite', 'lite', 'pt', 'pth', 'pkl', 'joblib', 'pbtxt', 'prototxt' ] }
+                { 
+                    name: 'All Model Files',
```
lutzroeder: Undo this change. File extensions should not be `xml` or `dot` as this will lead users to load generic files causing files.

lutzroeder:
General thoughts:
* Code should be simplified into fewer files following the patterns used for other frameworks.
* Remove `loadash` and edits in `setup.py`, `app.js` not related to adding OpenVINO support  
* Are there OpenVINO files available?
* Why are there multiple new formats needed and what's the difference between those formats? Can a single existing formats or standard be used instead?
* Try avoid using generic file extensions that will lead to conflicts and errors. For example, `.json` and `.pb` are causing conflicts, specific extensions like `.caffemodel`, `.mlmodel` or `.onnx` are preferable.
* Who will be using these tools and what are the user numbers? Are there clear benefits that warrant the added overhead...

src/openvino-dot/openvino-dot-metadata.js
```diff
@@ -0,0 +1,189 @@
...
+            results.push({
+                name: name,
+                connections: [{id: layerName}]
+            });
+            index++;
+        });
+        return results;
+    }
+
+    getAttributeVisible(operator, name, value) {
```
lutzroeder: This got moved to the .Attribute classes for most formats as it's only called once, e.g. try to do the JSON reading in the calling code.

src/openvino-dot/openvino-dot-metadata.js
```diff
@@ -0,0 +1,189 @@
...
+                }
+                if (attribute.hasOwnProperty('default')) {
+                    return value != attribute.default.toString();
+                }
+            }
+        }
+        return true;
+    }
+
+    static isEquivalent(a, b) {
```
lutzroeder: This got removed for most frameworks. Check if it's actually needed.

demid5111:
@lutzroeder general answer is the following. OpenVINO Toolkit is the unified and unique toolkit that provides the API for inference on edge devices, servers and other machines. There are options to infer on FPGA, Neural Compute Stick of Movidius etc. This is the only way for not-experts in hardware to deploy their models and have optimized inference there. Take a look at benchmark if interested (up to 19x times faster inference compared to frameworks). 

There are huge announcements about the tool wins in China, Europe, USA. Take a look [here](https://www.eetasia.com/news/article/18073102-intel-ready-for-ai-revolution) or [here](https://siliconangle.com/2018/09/04/baidu-intel-team-deliver-new-ai-services-china/)
This is why there is a huge need in supporting visualization of the model. 

Product exists for a long time and has multiple customers so there is no option to transition to another formats: so .xml and .bin can not be changed at all. Dot files can have a more particular extension, e.g. `filename.ovino.dot` or whatever.

src/openvino-dot/openvino-dot-metadata.js
demid5111: ok, sure, will check

src/openvino-dot/openvino-dot-metadata.js
demid5111: ok, will check

src/app.js
demid5111: please read the general comment below

demid5111:
@lutzroeder as for the OpenVINO models - yes they are available for download free of charge

electron-builder.yml
lutzroeder: Either way the change in `electron-builder.yml` should be reverted (same as `json` for example). Those extensions are too general and more likely than not `xml`, `bin` and `dot` files are not OpenVINO files and should not be opened in Netron.

package.json
lutzroeder: `filter` and `forEach` are available in JavaScript. It hasn't been used for any of the other formats. Your code is adding maintenance/dependency issues that didn't exist before.

setup.py
lutzroeder: Currently no strong compliance. Either way such formatting changes should be discussed and done consistently across files, not rolled into other PRs.

lutzroeder:
> @lutzroeder as for the OpenVINO models - yes they are available for download free of charge

Can you provide links to actual model files to download. 

Also can you provide some context what the difference between `bin`, `xml` and `dot` files are?

package.json
demid5111: @lutzroeder I have made the removal of dependency. The usage was a bit broader: filter, map, find, includes. However, there are alternatives in vanilla JS.

Are you going to resolve discussions that are fixed or I should do that?

demid5111:
@lutzroeder let ,e explain the difference between files via analogy with Caffe:

xml ~ protototxt
bin ~ caffemodel
dot ~ prototxt but for the runtime representation of the model

electron-builder.yml
demid5111: @lutzroeder could you please advice how to properly register those extensions, where is the `json` example?

src/app.js
demid5111: formatting is back to the original.

@lutzroeder is the discussion resolved now?

setup.py
demid5111: @lutzroeder I have removed new formatting

is the discussion resolved now?

demid5111:
@lutzroeder this is [the link to the models](https://download.01.org/openvinotoolkit/2018_R3/open_model_zoo/)

src/openvino-dot/openvino-dot-metadata.js
demid5111: @lutzroeder it was not used, I completely removed

is the discussion resolved now?

src/openvino-dot/openvino-dot-metadata.js
demid5111: @lutzroeder this is another thing I need you to help with. could you provide an example or explain a bit more what you expect me to do?

electron-builder.yml
lutzroeder: They are not registered and user will have to open the app and drag & drop or use the file dialog. This isn't ideal but there has been pushback from users and partners to not register general extensions. Even `pb` is at risk of getting removed.

setup.py
lutzroeder: Can you also remove the addition in `long_description`. This is not as complete as support for other frameworks and would also like to do some testing and see some user testing before adding it.

src/openvino-dot/openvino-dot-metadata.js
lutzroeder: Maybe check the code for some of the other frameworks like ONNX. The OpenVino code doesn't load any metadata in the 'OperatorMetadata' constructor implementation and won't do anything. For most other frameworks there is an `onnx-metadata.json` file which contains metadata such a the names for inputs (or the weight tensors which currently seems to be not showing in OpenVINO), coloring of operators via groups, documentation strings or which attributes should not show in the graph as they have default values. Maybe open a few `.onnx` files which have colors, show tensors and look more compact as they filter unnecessary attributes. Most of the `-metadata.json` files get updated via scripts under `./tools` when this is automatically possible but some data like the color groups needs to be manually edited.

lutzroeder:
> @lutzroeder this is [the link to the models](https://download.01.org/openvinotoolkit/2018_R3/open_model_zoo/)

Maybe add few of those files to `./test/model.json` with entries like this:

```
{
    "target": "openvino/age-gender-recognition-retail-0013.xml,openvino/age-gender-recognition-retail-0013.bin",
    "source": "https://download.01.org/openvinotoolkit/2018_R3/open_model_zoo/age-gender-recognition-retail-0013/FP32/age-gender-recognition-retail-0013.xml,https://download.01.org/openvinotoolkit/2018_R3/open_model_zoo/age-gender-recognition-retail-0013/FP32/age-gender-recognition-retail-0013.bin",
    "format": "OpenVINO IR",
    "link":   "https://download.01.org/openvinotoolkit"
  },
```

If you add them to the top of the file you can just run `npm run test` and they get loaded first. 

Might be also good to find other issues.

setup.py
demid5111: 1. I have joined all the files in two: `openvino-ir.js`, `openvino-ir-parser.js` and refactored the codebase so that the logic is massively reused across the targets. I think that we are done with it.

2. I do not think that the early check is needed. It is already performed in the `open` function by throwing the exception if not readable and that exception is gracefuly handled by us

3. Agree, that is the desired improvement that should be done after the this PR is merged. I will try to do it for sure in my next PR.

@lutzroeder is this discussion resolved now?

setup.py
demid5111: No problem, I have removed it. Please list what should be done in my next PR to make it possible for the OpenVINO to appear in the list of experimentally supported frameworks.

src/openvino-dot/openvino-dot-metadata.js
demid5111: @lutzroeder now I am not using the -metadata.json file, so that this comment is no longer actual, however, I added it to the list of future PRs.

can we close the discussion now?

setup.py
lutzroeder: 1. Getting close. Can you rename `openvino-ir.js` to just `openvino.js`, `openvino-ir-parser.js` and `openvino-parser.js` and `openvinoIR.ModelFactory` to `Openvino.ModelFactory`. You can use different namespace prefixes like `openvinoIR.Model` and `openvinoDot.Model` in the `openvino.js` file as they never need to be exported outside, only the unified `openvino.ModelFactory` is exported.

2. Basically any XML file the user opens that shows an error will be a bug on you? The unsupported check serves two purposes. a) filter out users loading random files from bugs b) if another framework supports XML files there might be another ModelFactory. If you can't handle a file it should be handed to the next factory. We can skip adding it for now but if usage will be any similar to JSON formats you will be dealing with this at some point.

src/openvino-dot/openvino-dot-metadata.js
lutzroeder: Remove the code maybe until it is being used. Otherwise it will have to be maintained and updated while it is not being used...

lutzroeder:
When done can you collapse the changes to a single commit.

```
git reset 8858cf0f710e90ae90d7171a321fdd23869d5da8 (or latest base commit)
git add *
git commit -m "Add OpenVINO support (#191)"
git push -f
```

electron-builder.yml
lutzroeder: This will get removed -- right?

src/openvino-ir.js
```diff
@@ -0,0 +1,656 @@
+/*jshint esversion: 6 */
+
+var openvinoIR = openvinoIR || {};
```
lutzroeder:
```
var openvino = openvino || {};
var openvinoIR = openvinoIR || {};
var openvinoDot = openvinoDot || {};
```

src/openvino-ir.js
```diff
@@ -0,0 +1,656 @@
...
+
+openvinoIR.ModelFactory = class {
```
lutzroeder: `openvino.ModelFactory = class {` only class that gets exported.

src/openvino-ir.js
```diff
@@ -0,0 +1,656 @@
...
+                if (attribute.hasOwnProperty('default')) {
+                    return value != attribute.default.toString();
+                }
+            }
+        }
+        return true;
+    }
+}
+
+openvinoIR.DotModel = class {
```
lutzroeder: `openvinoDot.Model`

src/view-browser.html
```diff
@@ -142,6 +142,7 @@ <h1 id='sidebar-title' class='sidebar-title'></h1>
 <script type='text/javascript' src='mxnet.js'></script>
 <script type='text/javascript' src='cntk.js'></script>
 <script type='text/javascript' src='sklearn.js'></script>
+<script type='text/javascript' src='openvino-ir.js'></script>
```
lutzroeder: `openvino.js`

src/view.js
```diff
@@ -18,6 +18,7 @@ var pytorch = pytorch || require('./pytorch');
 var sklearn = sklearn || require('./sklearn');
 var tf = tf || require('./tf');
 var tflite = tflite || require('./tflite');
+var openvinoIR = openvinoIR || require('./openvino-ir');
```
lutzroeder: `var openvino = openvino || require('./openvino');`

src/view.js
```diff
@@ -967,7 +968,8 @@ view.ModelFactoryService = class {
             new tflite.ModelFactory(),
             new tf.ModelFactory(),
             new sklearn.ModelFactory(),
-            new cntk.ModelFactory()
+            new cntk.ModelFactory(),
+            new openvinoIR.ModelFactory()
```
lutzroeder: `new openvino.ModelFactory()`

setup.py
```diff
@@ -118,7 +118,8 @@ def package_version():
             'pytorch.js', 'pytorch-metadata.json', 'pickle.js',
             'sklearn.js', 'sklearn-metadata.json',
             'tf.js', 'tf-metadata.json', 'tf-proto.js', 
-            'tflite.js', 'tflite-metadata.json', 'tflite-schema.js', 
+            'tflite.js', 'tflite-metadata.json', 'tflite-schema.js',
+            'openvino-ir.js', 'openvino-ir-parser.js',
```
lutzroeder: `openvino.js`, `openvino-parser.js`

src/openvino-ir.js
```diff
@@ -0,0 +1,656 @@
...
+    get nodes() {
+        return this._nodes;
+    }
+
+    get operators() {
+        return this._operators;
+    }
+}
+
+openvinoIR.DotNode = class extends openvinoIR.AbstractNode {
```
lutzroeder: `openvinoDot.Node` or even `openvino.dot.Node`

src/openvino-ir.js
```diff
@@ -0,0 +1,656 @@
...
+        this._inputs.push(id);
+    }
+
+    updateOutputs(id) {
+        this._outputs.push(id);
+    }
+}
+
+if (typeof module !== 'undefined' && typeof module.exports === 'object') {
+    module.exports.ModelFactory = openvinoIR.ModelFactory;
```
lutzroeder: `module.exports.ModelFactory = openvino.ModelFactory;`

test/models.json
```diff
@@ -2559,5 +2559,12 @@
     "source": "https://raw.githubusercontent.com/kosslab-kr/Tizen-NN-Runtime/master/Xor/xorGate.lite",
     "format": "TensorFlow Lite v3",
     "link":   "https://github.com/kosslab-kr/Tizen-NN-Runtime"
+  },
+  {
+    "type":   "openvino",
```
lutzroeder: To run `npm run test openvino` you will need to rebase to the latest `origin/master`. The code for running this was changed upstream.

src/openvino-ir.js
```diff
@@ -0,0 +1,656 @@
...
+
+var openvinoIR = openvinoIR || {};
+
+openvinoIR.ModelFactory = class {
+    match(context) {
+        return context.identifier.endsWith('.xml') || context.identifier.endsWith('.dot');
+    }
+
+    open(context, host, callback) {
+        host.require('./openvino-ir-parser', (err) => {
```
lutzroeder: Have a look at recent `tflite.js` for similar pattern. Attaching to `window` will not work in test environment.

```
host.require('./openvino-parser', (err, openvino_parser) => {
   var x = new openvino_parser.OpenVINOIRParser(...);
});

File: src/openvino-dot/openvino-dot-metadata.js (hunk 6 above)
Comment (demid5111) [reply src/openvino-dot/openvino-dot-metadata.js]: @lutzroeder I removed the method and the usage.

is the discussion resolved now?

File: electron-builder.yml (hunk 1 above)
Comment (demid5111) [reply electron-builder.yml]: @lutzroeder I removed the usage.

is the discussion resolved now?

File: src/openvino-ir.js (hunk 10 above)
Comment (demid5111) [reply src/openvino-ir.js]: @lutzroeder I made renaming and it seems already enough, not needed to make the sub namespace. 

is the discussion resolved now?

File: src/view-browser.html (hunk 11 above)
Comment (demid5111) [reply src/view-browser.html]: renamed

is the discussion resolved now?

File: src/view.js (hunk 12 above)
Comment (demid5111) [reply src/view.js]: renamed

is the discussion resolved now?

File: src/view.js (hunk 13 above)
Comment (demid5111) [reply src/view.js]: renamed

is the discussion resolved now?

File: setup.py (hunk 14 above)
Comment (demid5111) [reply setup.py]: renamed

is the discussion resolved now?

File: src/openvino-ir.js (hunk 16 above)
Comment (demid5111) [reply src/openvino-ir.js]: renamed

is the discussion resolved now?

File: src/openvino-ir.js (hunk 18 above)
Comment (demid5111) [reply src/openvino-ir.js]: I have started refactoring on my branch in the fork `feature/demidovs/fix-e2e-test`, however, when I run `npm run test openvino`, I get the following error:

```sh
events.js:307
    throw new TypeError('"listener" argument must be a function');
    ^

TypeError: "listener" argument must be a function
```

demid5111: @lutzroeder is it a known problem? I am now trying to rootcause it also

lutzroeder:
Manually merged as a single commit. Probably easiest to delete/recreate your fork from scratch.

demid5111:
@lutzroeder thank you! I am now starting to fill other gaps from the issue - it will be reflected in coming PRs

Issue #191, opened 2018-11-22
OpenVINO support


lutzroeder:
- [x] 1. Opening `rm_lstm4f.xml` results in TypeError (#192)
- [x] 2. dot files are not opened any more - need to fix it (#192)
- [x] 3. add preflight check for invalid xml and dot content
- [x] 6. Add test files to `./test/models.json` (#195) (#211)
- [x] 9. Add support for the version 3 of IR (#196)
- [x] 10. Category color support (#203)
- [x] 11. `-metadata.json` for coloring, documentation and attribute default filtering (#203).
- [x] 5. Filter attribute defaults based on `-metadata.json` to show fewer attributes in the graph
- [ ] 7. Show weight tensors
- [x] 8. Graph inputs and outputs should be exposed as Graph.inputs and Graph.outputs
- [x] 12. Move to `DOMParser`
- [x] 13. Remove `dot` support

demid5111:
Also leftovers from the PR 189:

- Put OpenVINO XML/DOT parser generation logic in the ./tools directory.
- Add support for the -metadata.json file for better filtering of properties, group coloring etc.

lutzroeder:
* Add `.dot` test files to `./test/models.json`

lutzroeder:
Opening `rm_lstm4f.xml` results in `TypeError: Cannot read property 'output' of undefined`.

lutzroeder:
* Graph inputs and outputs should be exposed as `Graph.inputs` and `Graph.outpus`

demid5111:
- Add support for the version 3 of IR

demid5111:
- dot files are not opened any more - need to fix it

demid5111:
- add preflight check for the content, so that xml and dot files that are not IR will not be opened by the openvino.ModelFactory

demid5111:
@lutzroeder I enumerated the list of issues to quickly reference in the PRs and also for discussions (status for the ww47.6):

1. Category color support
2. Filter attribute defaults to show fewer attributes in the graph
3. Show weight tensors
4. Put OpenVINO XML/DOT parser generation logic in the ./tools directory
5. Add support for the `-metadata.json` file for better filtering of properties, group coloring etc.
6. Add .dot test files to ./test/models.json
7. Opening `rm_lstm4f.xml` results in `TypeError: Cannot read property 'output' of undefined`. (#192)
8. Graph inputs and outputs should be exposed as `Graph.inputs` and `Graph.outpus`
[...]

PR #194, opened 2018-11-26
[OpenVINO] Added generation of OpenVINO parsers

demid5111:
This is the review for item !4 from #191 

In particular the opportunity to generate OpenVINO parsers from scratch if they are lost or the grammar was changed

### DIFF
--- a/tools/openvino
+++ b/tools/openvino
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+set -e
+
+root=$(cd $(dirname ${0}); pwd)
+src_dir=$(cd $(dirname ${0})/../src; pwd)
+ending="
+\nif (typeof module !== 'undefined' && typeof module.exports === 'object') {
+\n\tmodule.exports.IrParser = IrParser;
+\n\tmodule.exports.DotParser = DotParser;
+\n}"
+
+
+identifier=pytorch
+virtualenv=${build}/virtualenv/${identifier}
+
+if [ $(which python3) ] && [ $(which pip3) ]; then
+    python="python3"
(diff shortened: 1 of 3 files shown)


lutzroeder:
Looking at this some more, since `xml` is a regular XML file using a parser generator might be the wrong approach? Is it possible to use `DOMParser` instead to parse these files and read the DOM? We can look into using the `xmldoc` package in node, not requiring the large generated file and translation.

```
var fs = require('fs');
var text = fs.readFileSync('rm_cnn4a.xml', 'utf-8');

var xmldom = require('xmldom');
var parser = new xmldom.DOMParser();
var doc = parser.parseFromString(text, "application/xml");
```

demid5111:
> How is OpenVINO parsing these files? Is there an OpenVINO repository that has the grammar files checked in since all of OpenVINO must depend on them? Why did OpenVINO define a low-level grammar for an XML format?

this is the good question, however I am afraid I can not give an answer that you expect. Historically, the model format (IR) was never developed as a standard and therefore there is no such scheme. Actually, the runtime reader is implemented in C++, there is a generator of that format in Python and so on. The grammar was an attempt to parse XML quickly on the JS side. So, there is no any repo with official scheme.

My suggestion is:

1. I will add support for IR v3 as it is currently the new version introduced with the OpenVINO R4 release last month
2. I will work then on trying to transition to the common XML parser

Would you add the item in the #191 ?

lutzroeder:
Beware this is one of the reasons there will likely be more pushback accepting changes to generated parser code going forward. It bloats the repo with large diffs.

PR #195, opened 2018-11-27
Added dot models to testing

demid5111:
@lutzroeder this PR covers !6 from #191 As you suggested I use files attached to the gihub issue

### DIFF
--- a/test/models.json
+++ b/test/models.json
@@ -2323,6 +2323,27 @@
     "link":   "https://download.01.org/openvinotoolkit"
   },
   {
+    "type":   "openvino",
+    "target": "i8_normalized_squeezenet_ssd.dot",
+    "source": "https://github.com/lutzroeder/netron/files/2618902/i8_normalized_squeezenet_ssd.dot.zip[i8_normalized_squeezenet_ssd.dot]",
+    "format": "OpenVINO IR Dot",
+    "link":   "https://download.01.org/openvinotoolkit"
+  },
+  {
+    "type":   "openvino",
+    "target": "i8_normalized_squeezenet11.dot",
+    "source": "https://github.com/lutzroeder/netron/files/2618904/i8_normalized_squeezenet11.dot.zip[i8_normalized_squeezenet11.dot]",
+    "format": "OpenVINO IR Dot",
+    "link":   "https://download.01.org/openvinotoolkit"
+  },
+  {
+    "type":   "openvino",
+    "target": "i8_normalized_vgg16.dot",
+    "source": "https://github.com/lutzroeder/netron/files/2618905/i8_normalized_vgg16.dot.zip[i8_normalized_vgg16.dot]",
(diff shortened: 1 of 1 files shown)


lutzroeder:
Looking at this some more, `dot` files seem to be regular graphviz dot files and should not be loaded or rendered in Netron. The format is not an ML format that can be navigated but a general graph format for graphviz tools. Is there any advantage of having an interactive tool over running the file through dot and creating and image? I can make the commit to remove `dot` support which would also simplify things a lot as discussed earlier.

demid5111:
@lutzroeder Absolutely, there is a need in making dot rendered in Netron. Firstly, it is a business need to be able to render both formats (pre-inference and runtime versions of models) in one place in consistent UX. Our UX experts strongly consider this as a requirement. Secondly, we are going to show attributes, names etc. in a clear format - in the graphviz tools it is just a string that is hard to read, filter and analyze. So, I mean that graphviz renderers also loose in the interactivity compared to what Netron provides.

that was the motivation. however, as dot files are now supported, adding tests to make sure everything works correctly is essential to keep product functional

lutzroeder:
Maybe we should walk through this by example as I'm still not sure I understand how this is being used. How are `dot` files generated and why did you choose to "store" and "run" a machine learning model in a format that is meant for graphical representation? These files include attributes like `shape: box` or `style: filled` or `fillcolor: #20F608`?

demid5111:
the format allows you to have a render engine - graphviz, and this is the main reason the format was selected, however such representations loose domain specificity (like there are layers with properties) and interactivity. the model is never run from that format, however the intention is another. Imagine you have a model, trained model. Then you want it to appear in the application. You write the app, infer the model there and get performance, say 10fps. It is accurate but not quick. What if you had an option to get 100fps? With OpenVINO you can do it, BUT for that you need to create IR (.xml + .bin) from the trained model. OK, now you run the app, it shows 100fps. Awesome. Then you try to understand - how did it become so quick? Can I do better? For that you need to understand what happens with you model during inference. The only way to do it, is to look at the .dot file with that model. There you will see that actually your convolutions and ReLUs were merged although in IR you see them as two different layers. And there are tons of such optimization that might happen. So the idea is to be able to see and explore the graph.

lutzroeder:
Ideally wouldn't want to set the expectation that this is a supported scenario that users rely on. Removing features once users rely on them can be painful. Do you have a timeline when this gets removed?

demid5111:
@lutzroeder looking at the list of other items that we defined in #191, I expect to make the parser probably by the end of January (if we talk realistically). However, maybe I overestimate the complexity of closing remaining tasks and I can start earlier. Anyway, when everything else is closed I will proceed with XML parser and replacing dot with another xml in order to keep the format unified.

demid5111:
@lutzroeder thanks for the merge

PR #196, opened 2018-11-28
[OpenVINO] Add v3 IR support

demid5111:
This PR covers !9 from #191 

In particular, IR v3 compared to IR v2 is support of new Layer - TensorIterator with the complex internal structure


lutzroeder:
Since the diff for `openvino-parser.js` is large and will bloat the repro we should try and minimize making changes to this file. Are you sure the current version is finally and will need no further edits? Otherwise, can you do some testing and add any other changes so this file doesn't need to be regenerated or updated again. Alternatively, switching to `DOMParser` and `xmldom` sooner would allow for some more flexibility.

demid5111:
@lutzroeder I confirm, this is the latest major change for the parser before transitioning to more robust parser libraries. please merge it if you have no other questions

demid5111:
@lutzroeder thank you, now switching to the `.bin` content processing and adding the `-metadata.json` support for OpenVINO

PR #203, opened 2018-12-13
[OpenVINO] Support per-layer documentation 

demid5111:
This PR covers items !10,11 from #191


lutzroeder:
`openvino.js` was missing the `marked` import which would fail in the Electron app. Also `getOperatorDocumentation` has been moved to `openvino.Node::documentation` so the code was there already. Merged as 0fc47767d8890086217b68fc48fa81319cef1110.

PR #230, opened 2019-02-15
Add back pass for server.ThreadedHTTPServer

chiragjn:
https://github.com/lutzroeder/netron/blob/66d68bd74c16dfd568ec4df5cb6ef801f92e435c/src/server.py#L112-L114

A `pass` was removed (possibly by mistake) 9 days ago which causes the netron==2.8.7 server to break due to IndentationError
```python
    class HTTPServerThread(threading.Thread):
        ^
IndentationError: expected an indented block
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,12 @@
     "repository": "lutzroeder/netron",
     "main": "src/app.js",
     "scripts": {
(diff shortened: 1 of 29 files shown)


lutzroeder:
#232

PR #231, opened 2019-02-17
[OpenVINO] Adding support for TensorIterator and improved validation (#191)

demid5111:
@lutzroeder  this PR covers:

1. Fixing the error with the TensorIterator after the transition to DOM parser instead of PEG.js parser
2. Added more test cases that reveal the problem that was not caught after the transition to DOM parser
3. Added self-validation of the OpenVINO graph: when the graph is built we need to make sure that there is only one connected component. Earlier, existing tests did not check it resulting in the Netron producing multiple connected components for the OpenVINO IR which is not correct by definition.


demid5111:
@lutzroeder thank you for proactively merging it. I am switching now to the weights loading

PR #234, opened 2019-02-26
[OpenVINO] Integrating Netron as a component to another tool

demid5111:
@lutzroeder in OpeVINO there is a need now to be able to visualize the graph in some internal tools. I have made everything and it works with only small number of tweaks with the Netron code base. 

However, I feel that keeping my own fork is not  scalable as the changes that I need in Netron are really minor - exposing one more class in the namespace and catching import error.


lutzroeder:
Can you provide some context why these changes are needed and how they are used?

lutzroeder:
Can you explain how this is used? Why is the `try {} catch {}` needed and why does `BrowserContext` need to be exposed?

demid5111:
@lutzroeder those changes are needed as I wrap Netron into Angular 7 component and provide OpenVINO IR directly instead of making users upload it to the service. This requires me to trigger openning of the model directly and therefore when I call it, i need a context, that I create manually and I need this class to be available. As for the try/except, Angular applications use webpack system and when importing the file, it throws exception. Therefore I need that catch.

How harmful proposed changes are from your perspective?

lutzroeder:
`BrowserContext` implements an interface, it should be possible to have the client code copy this or have a custom implementation? The `try {} catch {}` is more problematic. There are different ways in which this code is hosted and exceptions should not be muted as it will make it harder to find real issues. Why exactly are exceptions thrown and can they be fixed in another way?

demid5111:
@lutzroeder the exception happens due to the fact that the `View` is not available to the time when the script is executed. As I said, this is wrapped into the Angular component with webpack-controlled imports without dynamic requires due to security threats that this dynamic import policy imposes. At the same time, I did not find a way to obtain `BrowserContext` in a proper way as it is later used by the code that I do not control from the Angular application - in Netron core sources. How bad is it from your standpoint to expose `BrowserContext`?

lutzroeder:
For both changes still trying to understand if there isn't a different way to approach this. Maybe it is possible to share a code example including webpack packaging? Catching all exceptions is definitely a no as it will mute real failures as well. `Context` and `Host` are both interfaces so it should be possible to copy or reimplement them without adding more module dependencies.

demid5111:
@lutzroeder give me time till end of week, I will try to share the sample code with you.

Issue #168, opened 2018-10-25
Loop, If and Scan support


dshirron:
Is there roadmap for supporting operators of loop type:
For example:
 ONNX Loop
Caffe2 ONNXWhile

lutzroeder:
Can you share a sample file for each and describe what support you are looking for?

dshirron:
`from caffe2.python import workspace, model_helper,core
import numpy as np
from caffe2.proto import caffe2_pb2

# Now import the caffe2 mobile exporter
from caffe2.python.predictor import mobile_exporter

from caffe2.python.onnx import frontend
import onnx

# Create the initial input data
workspace.ResetWorkspace()
max_trip_count = np.full(1,20).astype(np.int64)
condition = np.full(1,True).astype(np.bool)
first_init = np.full((1),1).astype(np.float32)
second_init = np.full((1),1).astype(np.float32)
workspace.FeedBlob("max_trip_count", max_trip_count)
workspace.FeedBlob("condition", condition)
workspace.FeedBlob("first_init", first_init)
workspace.FeedBlob("second_init", second_init)

# Create body net
body_net = caffe2_pb2.NetDef()
# Two loop carried dependencies: first and second
body_net.external_input.extend(['i', 'cond', 'first', 'second'])
body_net.external_output.extend(['cond_new', 'second', 'third', 'third','cond','cond'])
add_op = core.CreateOperator(
    'Add',
    ['first', 'second'],
    ['third'],
)
print_cond = core.CreateOperator(
    'Print',
    ['cond'],
    [],
)
print3 = core.CreateOperator(
    'Print',
    ['third'],
    [],
)
limit_const = core.CreateOperator(
    'ConstantFill',
    [],
    ['limit_const'],
    shape=[1],
    dtype=caffe2_pb2.TensorProto.FLOAT,
    value=1000.0,
)
cond = core.CreateOperator(
    'LT',
    ['third', 'limit_const'],
    ['cond_new'],
    broadcast=1
)
body_net.op.extend([add_op, print_cond,print3, limit_const, cond])

while_op = core.CreateOperator(
    'ONNXWhile',
    ['max_trip_count', 'condition', 'first_init', 'second_init'],
    ['first_b', 'second_a', 'kabiba','kusa','asd'],
    body=body_net,
    has_cond=True,
    has_trip_count=True,
    save_scopes=0,
)

main_net = caffe2_pb2.NetDef()
main_net.op.extend([while_op])
[...]

Issue #68, opened 2018-01-24
Hierarchical expansion of large graphs


jenniferzhu:
I failed to load 2.1 MB .meta file generated from char-rnn model (https://github.com/sherjilozair/char-rnn-tensorflow) with python 2.7 and tensorflow 1.0.

It is loading forever... I am wondering what is wrong with these files?

lutzroeder:
How did you produce the `.meta` file? Can you share it?

lutzroeder:
The graph has more than 5000 nodes and the graph layout is choking. Tried running GraphViz and it's running forever as well. Those graphs will need an expand and explore UI that avoids rendering the entire graph. TensorFlow support is pretty experimental at this point.

jenniferzhu:
Thanks @lutzroeder 
I hope there would be some ways to workaround this issue, since I really like Netron's image of models.

erwincoumans:
Same issue, meta file is here in case you want to test it:
https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_data/policies/ppo/minitaur_reactive_env/model.ckpt-14000000.meta

srikanthdevalour:
Netron doesnot show FC layer in the diagram. I need to view the weights of the FC layer. Is there anything we can do it view Fully Connected layer.

lutzroeder:
@srikanthdevalour can you create a new issue and add a sample file.

kobemini:
@lutzroeder You mentioned "Those graphs will need an expand and explore UI that avoids rendering the entire graph" , are there any improvements planned for release versions in the future?

jojivk73:
For large graphs, is it possible to add an option to show a given node +  fanin and fanouts upto specified level.
Example netron <g.pbtxt> -node <node_name> -level 3

JasonWayne:
@jojivk73  there is no such args like node or level in my netron cmd line...

yairb-gm:
Hi, I'm encountering this issue as well for models I was able to open with previous versions, of onnx files

lutzroeder:
@yairb-gm can you share a sample file to reproduce the issue.

Fannettembrown:
Also check an see if your branch system has a Flutter system....

CarlPoirier:
Hi @lutzroeder,

[...]

PR #318, opened 2019-08-09
Add basic tensorflow subgraph (function) support

scottcjt:
Hi,

This patch adds basic `FunctionDef` support in TF. Basically a `FunctionDef` is a `GraphDef` with some minor differences such as:

* graph inputs/outputs are explicitly declared
* when referencing a node output, the format is `<node_name>:<output_arg_name>:<port>`. The **output_arg_name** is determined by op type, and it is hidden in this patch (which is not a good thing IMHO...).

### DIFF
--- a/src/tf.js
+++ b/src/tf.js
@@ -228,8 +228,23 @@ tf.Model = class {
             else if (model.meta_graphs.length > 1) {
                 name = i.toString();
             }
+            else {
+                name = 'main';
+            }
             this._graphs.push(new tf.Graph(metadata, metaGraph, name));
         }
+
+        // Recursively add all subgraphs.
+        let visited_graph = [];
+        let pending_graphs = [...this._graphs];
+        while (pending_graphs.length > 0) {
+            let g = pending_graphs.shift();
+            visited_graph.push(g);
+            for (let f of g.functions)
+                pending_graphs.push(f);
+        }
+        this._graphs = visited_graph;
+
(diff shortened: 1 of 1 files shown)


lutzroeder:
@scottcjt Can you provide a sample file to test this change?

scottcjt:
Sure. But I don't know any public url-reachable model using this, so here is a sample graph. [tf_function.tar.gz](https://github.com/lutzroeder/netron/files/3491480/tf_function.tar.gz)

The most easy way to get a TF graph with functions is to use TF-2.0 and its new control flow. Sample above is generated by:

```python
import tensorflow as tf

def save_to_pb_and_txt(name, gdef):
    from google.protobuf.text_format import MessageToString

    with open('%s.pb' % name, mode='wb') as fp:
        fp.write(gdef.SerializeToString())

    with open('%s.pbtxt' % name, mode='w') as fp:
        fp.write(MessageToString(gdef))

def build_simple_loop_graph():
    @tf.function(input_signature=[tf.TensorSpec(shape=[1], dtype=tf.int32)])
    def main(n):
        total = 0
        i = 0
        while i < n[0]:
            total += i
            i += 1

        return tf.identity(total, name='out')

    # Test...
    result = main([10])
    assert result.numpy() == sum(range(10)), 'expect %d, got %d' % (sum(range(10)), result)

    cfunc = main.get_concrete_function()
    save_to_pb_and_txt('xd', cfunc.graph.as_graph_def())

build_simple_loop_graph()
```

lutzroeder:
#168 #68

PR #340, opened 2019-10-05
Visualization support for MNN models

SatanWoo:
**MNN is a lightweight and efficient deep neural network inference engine developed by Alibaba Group, which can be referred from following link:**

[MNN](https://github.com/alibaba/MNN)

For this pull request, we have implemented all necessary visualization functions, for example:
- load .mnn models.
- present all op-related parameters in the sidebar.
- visualize the entire model as a graph

Since netron is a widely-used, we hope to contribute ourselves to integrate MNN functionalities into it.


SatanWoo:
I think the implementation of `setup.py` has additional **','** 

I don't think I need to resolve this conflict.

SatanWoo:
Ok, finally I have cleared all conflicts. The behavior looks very weird.

PR #345, opened 2019-10-07
Fix loading .mnn model failed when there is no extraTensorDescribe existed.

SatanWoo:
Hi @lutzroeder 

I'm very appreciated with your refactored code for supporting MNN visualization. However, there is a minor bug where the `net.extraTensorDescribe` **does not** always exist. So under that circumstance will `Netron` fail to load `.mnn` models.

In this pull request, I fix this bug and attach a `simple.mnn` file for you to easily reproduce the bug.

[simple.mnn.zip](https://github.com/lutzroeder/netron/files/3698308/simple.mnn.zip)


lutzroeder:
[simple.mnn.zip](https://github.com/lutzroeder/netron/files/5405960/simple.mnn.zip)

PR #363, opened 2019-10-31
Support armnn serialized format

Tee0125:
here's patch to support armnn format. (experimental)

armnn-schema.js is compiled from ArmnnSchema.fbs included in armNN serailizer.

see also:
* armnn: https://github.com/ARM-software/armnn

PR #364, opened 2019-10-31
add armnn serialized format support

Tee0125:
here's patch to support armnn format. (experimental)

armnn-schema.js is compiled from ArmnnSchema.fbs included in armNN serailizer.

see also:

armnn: https://github.com/ARM-software/armnn

As mensioned in #363, I will check items in below:

- [x] Add sample files to test/models.json and run node test/test.js armnn
- [x]  Add tools/armnn script and sync, schema to automate regenerating armnn-schema.js
- [x]  Add tools/armnn script to run as part of ./Makefile
- [x]  Run make lint


Tee0125:
```Can you provide some context what armnn is and who is using it? Why is there another format needed given there are a dozen formats already? Are there any .armnn files available?```

ArmNN is Arm's NN framework, and is one of backend of Android NN API. So companies who wan to support Android NN API with less effort and have NN accelerating chip will be using it.

Since I'm not employer of Arm, I have no idea why ArmNN developers defined a new format. But as I reviewed protobuf with tensorflow scheme and onnx scheme were not good enough to describe quantized tensor.

with ArmNN converter included in ArmNN, onnx/pb/caffe files can be easily converted to .armnn file and I requested ArmNN developers to upload some sample armnn files.

File: src/app.js [hunk 1]
```diff
@@ -131,16 +131,18 @@ class Application {
                     'pt', 'pth', 't7',
                     'pkl', 'joblib',
                     'pbtxt', 'prototxt',
+                    'armnn',
                     'cfg',
                     'xml' ] }
             ]
         };
-        const selectedFiles = electron.dialog.showOpenDialogSync(showOpenDialogOptions);
```
lutzroeder: Looks like this failed to merge?

tools/armnn-script.js
```diff
@@ -0,0 +1,17 @@
+var flatbuffers = require('../third_party/flatbuffers/js/flatbuffers.js').flatbuffers;
```
lutzroeder: Can you remove this script. It shouldn't be needed and same can be accomplished in `armnn.js`.

tools/armnn
```diff
@@ -0,0 +1,73 @@
...
+    pushd "${third_party}/flatbuffers" > /dev/null
+    cmake -G "Unix Makefiles" . > /dev/null
+    make > /dev/null
+    popd > /dev/null
+}
+
+schema() {
+    bold "armnn schema"
+
+    # master branch is patched not to make problem with js but current release branch is not patched yet
```
lutzroeder: Would it make sense to clone the `master` branch?

src/armnn.js
```diff
@@ -0,0 +1,625 @@
...
+    }
+
+    get visible() {
+        return this._visible == false ? false : true;
+    }
+};
+
+armnn.Parameter = class {
+
+    constructor(name, id, tensorInfo, initializer) {
```
lutzroeder: Should take `args` like the other implementations. `Argument` instances should get generated in the code that is calling the constructor.

src/armnn.js
```diff
@@ -0,0 +1,625 @@
...
+
+    get visible() {
+        return true;
+    }
+
+    get arguments() {
+        return this._arguments;
+    }
+
+    static makeKey(layer_id, index) {
```
lutzroeder: `makeKey` is only used by `Graph` so it should move there and is private `_makeKey`.

src/app.js
Tee0125: sorry for confusing, this is mis-pushed one.

commit f6a3411 makes problem with opening file in my Mac OS X, I revert the commit for testing. and it should not be pushed.

src/armnn.js
Tee0125: fixed

tools/armnn
Tee0125: fixed

lutzroeder:
The test file `ssd_mobilenet_v3_small_coco_2019_08_14.armnn` renders empty when opened?

You are still working on removing the `tools/arm-*.js` file -- correct?

Tee0125:
pushed armnn-schema.js was a problem. After regenerating it with 'tools/armnn sync install schema', graph is shown correctly.

(while doing that, I found tools/armnn not working, so I fixed it either)

Since removing unnecessary require statement in `./tools/armnn-script.js` was already done, I think, there's nothing left to do.

lutzroeder:
Regarding `./tools/armnn-script.js` the question was if we remove that script entirely. Custom code generation tends to be difficult to maintain and this script shouldn't be needed. It should be possible to implement this in `src/armnn.js` directly?

Tee0125:
I had no idea whether is it possible or not.

Anyway I tried to implement it directly, and checked it's working. :)

lutzroeder:
1) `git clone https://github.com/openedges/netron.git && cd netron`
2) `npm install`
3) `node ./test/test.js armnn`
4) `npx electron .`
5) Open `./test/data/armnn/ssd_mobilenet_v3_small_coco_2019_08_14.armnn`

This shows an empty screen?

Tee0125:
in my OS X, it does not show empty graph.

screenshot

Tee0125:
model url was a problem. I confirmed `node test.js armnn` downloads raw armnn model and fresh cloned netron shows graph properly

lutzroeder:
Manually merged this PR. Can you check if master includes all the changes and recreate your fork?

Tee0125:
Please merge two more commits, ef1cd473ff553edbcf5ec6e8196c131b39938f1d and 7c8be62ee302684ca6008aa754a53a854904dae3

lutzroeder:
Those changes are included in d2849d6b18a58c7d125d79150aa3bc1cb1c2f327 -- right?

Tee0125:
those changes are included. however I found a neat (made while changing type name from lowercase to camel case) :'(

so one more patch( 42269b5ab8a9d17165fadaec35b9e5fa8e6ff2bc ) needs to be merged

Thanks

lutzroeder:
#515

PR #368, opened 2019-11-08
Update .gitignore

ken4204708:


lutzroeder:
Why is this change needed?

PR #369, opened 2019-11-12
bug fix on displaying quantized tensor

Tee0125:
fix typo

### DIFF
--- a/src/armnn.js
+++ b/src/armnn.js
@@ -551,7 +551,7 @@ armnn.TensorType = class {
     }
 
     isQuantized() {
-        return this._dataType.startsWith("quantised");
+        return this._dataType.startsWith("Quantised");
     }
 };
 


lutzroeder:
Is `QuantisedAsymm8` the same as `quint8` and `QuantisedSymm16` same as `qint16`?`Signed32` seems the same as `int32`?

Tee0125:
here are sample files. (In case of network in ..._q16.armnn, weight tensor is quantized to asymetric 8bit fixed point, bias is quantized to 32bit integer and activation is quantized to symmetric 16bit fixed point) 
 
[samples.zip](https://github.com/lutzroeder/netron/files/3839429/samples.zip)

Tee0125:
and I think, `qint8` should be `quint8`, and `quint16` should be `qint16`

lutzroeder:
#515

PR #371, opened 2019-11-13
[OpenVINO] Fixed layer precision retrieval for IR v7 and newer

demid5111:
This PR contains:

1. Fix for precision retrieval. Precision is a part of OpenVINO IR layers of IR v6 and earlier. In IR v7 and newer the port is no longer an attribute of the layer but instead is a part of each output port and blob. This PR handles this case while keeping the approach backward compatible.

2. Fix for TensorIterator extraction. In recent versions of Netron this logic was broken and the TensorIterator layer that is in nature just a container was shown as a single layer. Instead, there is a logic to extract its contents and replace each TensorIterator node with a complete subgraph that it contains.


lutzroeder:
Can you add a test file to reproduce these issues?

demid5111:
Yes, sure, that was my plan for tomorrow :) I will attach them to MR and include in our tests. среда, 13 ноября 2019г., 22:12 +03:00 от Lutz Roeder  notifications@github.com :

lutzroeder:
```
TypeError: Cannot read property '_inputs' of undefined
    at openvino.Graph._replaceTensorIteratorWithSubgraph (/Users/lutzroeder/Projects/netron/src/openvino.js:281:32)
    at new openvino.Graph (./netron/src/openvino.js:129:14)
    at new openvino.Model (./netron/src/openvino.js:70:21)
    at ./netron/src/openvino.js:55:24
```

PR #384, opened 2019-12-12
update darknet-metadata.json

HagegeR:
added missing net option, some descriptions, and some attributes type.

can you please help me on the TODO's and in the "from" attribute (a kind of routing) ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
         "flatbuffers": "1.11.0",
         "handlebars": "4.5.3",
         "long": "4.0.0",
-        "marked": "0.7.0",
+        "marked": "0.8.0",
         "pako": "1.0.10",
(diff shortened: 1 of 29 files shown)


lutzroeder:
@HagegeR can you update the PR to latest? `boolean` is now handled as `int32` in `master` and `from` is handled for  `scale_channels`.

HagegeR:
done, what about shortcut and sam? is from supported there too?

also I don't know why but it's as if I changed all the files although the files are the same (21 file change when it's only 2)

lutzroeder:
Can you validate the change? `node ./test/test.js darknet` fails with errors.

HagegeR:
can I see the errors?

lutzroeder:
Already fixed and manually merged to ‘master’.

PR #393, opened 2019-12-17
delete non existent layer normalization

HagegeR:
this just make every node bigger for no reason, batch_normalize is just a property


lutzroeder:
Most frameworks model batch normalization as a separate layer or an inner layer as it is a separate operation using weight tensors. It currently renders as an inner node for consistency.

HagegeR:
not in Darknet, in Darknet it's just an option that you turn on or off.

Issue #307, opened 2019-07-29
OpenVINO weights support


GlebKazantaev:
OpenVINO models (IRs) contains XML and BIN files. For now Netron loads only XML file with topological description. And it would be great if Netron will load weights from BIN file and display values as implemented in other frameworks (like in ONNX).
![image](https://user-images.githubusercontent.com/18436845/62047352-ee60fd80-b212-11e9-8c6c-dc597d4dad75.png)

All information about offsets, sizes and types of tensor can be found in XML.

Note: BIN files are optional an may not exists in the same folder with XML file.

lutzroeder:
Duplicate of #191. @demid5111

lutzroeder:
@GlebKazantaev @demid5111 What format are `.bin` files stored in? Is it possible to detect the `.bin` format from other `.bin` files which use the same extension but are not generated by OpenVINO?

lutzroeder:
@demid5111

PR #406, opened 2020-01-07
Update README.md

eladeyal-intel:


lutzroeder:
#307 was not implemented so `bin` is not supported.

eladeyal-intel:
I thought it was at least partly in.

Because of this https://github.com/lutzroeder/netron/commit/494e430e5f278246faf7ac43900e19d1afa1bc24

Then I tried downloading these files and open with desktop app

https://download.01.org/opencv/2019/open_model_zoo/R3/20190905_163000_models_bin/driver-action-recognition-adas-0002-decoder/FP32/driver-action-recognition-adas-0002-decoder.xml
https://download.01.org/opencv/2019/open_model_zoo/R3/20190905_163000_models_bin/driver-action-recognition-adas-0002-decoder/FP32/driver-action-recognition-adas-0002-decoder.bin


I can see the biases which I think come from the bin files.

image


image


If I open the xml  file via the web netron, where the bin is not available, it will not show these blobs.

@lutzroeder

PR #414, opened 2020-01-19
Force dark mode using the electron config file entry

tomdol:
I'm on Ubuntu 18.04 and for some reason the app ignores the system theme setting.

dark_mode
debugger

With this PR I'd like to propose a way to force the dark mode and ignore the system settings (which is what electron uses by default to determine which color mode should be used).

To acchieve it, I've editted the `configuration.json` file (located in `~/.config/Electron` in my case) and added an entry: `"forceDarkMode": true`. This value is then used to switch the `electron.nativeTheme.themeSource` property which controls the color scheme of the app https://electronjs.org/docs/api/native-theme#properties

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -92,6 +92,9 @@ class Application {
         if (!this._configuration.has('userId')) {
             this._configuration.set('userId', require('uuid').v4());
         }
+        if (this._configuration.get('forceDarkMode'))  {
+            electron.nativeTheme.themeSource = 'dark';
+        }
         global.userId = this._configuration.get('userId');
         if (this._openFileQueue) {
             let openFileQueue = this._openFileQueue;


lutzroeder:
Is there an Electron issue for this? Any reason this can’t be fixed in Electron?

tomdol:
The only related issue I found that has something to do with theming it is this one https://github.com/electron/electron/issues/19932. It's not a bugreport for a situation I have on my system though.

Nevertheless I would really like to have a way to force the dark (or light) mode in Netron and not depend on the system settings. Perhaps a "Dark mode" toggle or a submenu in the "View" dropdown (containing "system", "light" and "dark") would be an alternative to what I did?

lutzroeder:
Custom theming support beyond default OS behavior is not a priority right now. Understanding the Electron behavior on Linux would be helpful.

Issue #418, opened 2020-01-21
[OpenVINO] Sigmoid activation type is not shown


pavel-rodionov:
Even when "Show attributes" option was enabled `sigmoid` activation type wasn't show up:
<img width="1136" alt="Screenshot 2020-01-21 at 11 21 05" src="https://user-images.githubusercontent.com/55975500/72787801-a959d900-3c41-11ea-959c-93b9a1770129.png">

Please find example model in attachment:
[model.xml.zip](https://github.com/lutzroeder/netron/files/4089939/model.xml.zip)

Is seems like the issue is in `openvino.js` where value is equal to default schema value visibility of this element is disabled. PR with possible fix will be submitted.

lutzroeder:
'sigmoid' is the default value for the property. Defaults are not rendered in the graph.

PR #419, opened 2020-01-21
[OpenVino] Fix issue when Sigmoid activation is not shown

pavel-rodionov:
Even when "Show attributes" option was enabled sigmoid activation type wasn't show up.

Possible fix for https://github.com/lutzroeder/netron/issues/418


lutzroeder:
Is `sigmoid` not the correct default value for this attribute?

PR #482, opened 2020-05-02
size and offset in openvino blobs to numbers

ArutyunovG:
This is observed in IRv10, so it can probably be not supported.

`typeof(offset) == "string"` and `typeof(size) == "string".`
As a result, we get that "string" + "string" is compared with "number" 
`(offset + size) <= bin.length`
and the result might not be the one expected.

For my model it ended up loading weights only for the first few layers.


lutzroeder:
Can you share a sample file?

ArutyunovG:
The same as #483

PR #483, opened 2020-05-02
check if openvino precision is null

ArutyunovG:
The error solved is 
Error loading OpenVINO model. Cannot read property 'toLowerCase' of null in 'netname.xml'.
when trying to load xml file.

It is observed with IRv10.
precision null is handled in switch, so it's probably ok that it is passed.

### DIFF
--- a/src/openvino.js
+++ b/src/openvino.js
@@ -877,7 +877,10 @@ openvino.Tensor = class {
 openvino.TensorType = class {
 
     constructor(precision, shape) {
-        switch (precision.toLowerCase()) {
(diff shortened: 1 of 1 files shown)


lutzroeder:
Can you share a sample file?

ArutyunovG:
Added sample here https://dropmefiles.com/4sJze with zero weights, with exception of SSD head (reshape, transpose stuff), so it still can be used.
Just for clarification, 
parseInt(data['offset'], 20)
is probably intended as 
parseInt(data['offset'], 10)?

Issue #498, opened 2020-05-22
sklearn ColumnTransformer and FeatureUnion support


mbalatsko:
It would be great to add a feature of visualization branching inside `sklearn.pipeline.Pipeline`. For now, all transformers which are inside  `ColumnTransformer` or `FeatureUnion` are not visualized and are shown as attributes. This makes not possible to get pipeline structure insight. To be more clear I'll add an example how it should look like.

Current visualization:
![Current vizualization](https://user-images.githubusercontent.com/15967073/82681012-6c344080-9c4d-11ea-9e6b-4c1c0f1b4a0e.png)

Improved visualization:
![Improved visualization](https://user-images.githubusercontent.com/15967073/82681296-dea52080-9c4d-11ea-85eb-c388eeaddc87.png)

mbalatsko:
I can help you with this improvement and actually get it done. If you are interested do not hesitate to contact me. Your project is used in one product that I really like and I would really appreciate this improvement

lutzroeder:
@mbalatsko can you share a sample file?

mbalatsko:
Code with column transformer:
```python
from sklearn.pipeline import Pipeline, FeatureUnion
from sklearn.compose import ColumnTransformer
from sklearn.preprocessing import StandardScaler, OneHotEncoder
from sklearn.impute import SimpleImputer
from sklearn.linear_model import LogisticRegression
from sklearn.feature_selection import SelectKBest
from sklearn.feature_extraction.text import TfidfVectorizer


categorical_columns = ['Sex', 'JobType']
numeric_columns = ['Age', 'Salary']
text_columns = ['job_description']

model = Pipeline([
    ('preprocessing', ColumnTransformer([
        ('text_preprocessing', TfidfVectorizer(), text_columns),
        ('categorical_preprocessing', OneHotEncoder(), categorical_columns),
        ('numeric_preprocessing', Pipeline([('impute', SimpleImputer()), ('scale', StandardScaler())]), numeric_columns)
    ])),
    ('fs', SelectKBest(k=1000)),
    ('clf', LogisticRegression())
])
```
Model file:
[column_transformer_pipeline.zip](https://github.com/lutzroeder/netron/files/4670399/column_transformer_pipeline.zip)

mbalatsko:
Code with feature union:
```python
from sklearn.pipeline import Pipeline, FeatureUnion
from sklearn.compose import ColumnTransformer
from sklearn.decomposition import PCA, TruncatedSVD
[...]

PR #501, opened 2020-05-23
Add sklearn ColumnTransformer and FeatureUnion support (#498)

mbalatsko:
- Add support of visualization of inner transformers of `ColumnTransformer` and `Feature union`
- Add dataset graph nodes to be able to get info about `ColumnTransformer` transformers columns application. Also, add dataset graph nodes for data stream synchronization to improve model graph visualization (not to create m:n connection, if for example `ColumnTransformers` or `FeatureUnion` are applied in a row)

Visualization example
Code:
```python
categorical_columns = ['Sex', 'JobType']
numeric_columns = ['Age', 'Salary']
text_columns = ['job_description']

model = Pipeline([
    ('preprocessing1', FeatureUnion([
        ('pca', PCA()),
        ('svd', TruncatedSVD()),
    ])),
    ('preprocessing2', ColumnTransformer([
        ('text_preprocessing', TfidfVectorizer(), text_columns),
        ('categorical_preprocessing', OneHotEncoder(), categorical_columns),
        ('numeric_preprocessing', Pipeline([('impute', SimpleImputer()), ('scale', StandardScaler())]), numeric_columns)
    ])),
    ('preprocessing3', ColumnTransformer([
        ('text_preprocessing', TfidfVectorizer(), text_columns),
        ('categorical_preprocessing', OneHotEncoder(), categorical_columns),
        ('numeric_preprocessing', Pipeline([('impute', SimpleImputer()), ('scale', StandardScaler())]), numeric_columns)
    ])),
    ('fs', SelectKBest(k=1000)),
    ('clf', LogisticRegression())
])
```

Model file:
[pipeline.zip](https://github.com/lutzroeder/netron/files/4672307/column_pipeline.zip)

Output model graph:
Output model graph

### DIFF
--- a/src/sklearn.js
+++ b/src/sklearn.js
@@ -89,6 +89,14 @@ sklearn.Model = class {
     }
 };
 
+sklearn.Data = class {
+    constructor(name, columns = []) {
+        this.columns = columns;
+        this.__name__ = name;
+        this.__module__ = '';
+    }
+};
+
 sklearn.Graph = class {
 
     constructor(metadata, obj, weights) {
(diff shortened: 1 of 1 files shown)


lutzroeder:
Adding new nodes with made up type names is confusing. Could clusters be used instead similar to pipelines or would information be lost?

mbalatsko:
@lutzroeder What do you mean by clusters?

lutzroeder:
For nested pipelines in Core ML for example nodes are clustered under a parent using a group name. This might not work here as the information of the parent operation is lost. Basically the model has hierarchy and can't be rendered as a flat graph. The right approach seems to be to have special expandable nodes that can render subgraphs?

screenshot

PR #512, opened 2020-06-08
add tnn model support

masterwan96:
I adapt tnn's network. You can check it


lutzroeder:
@nihui can you provide some context? Why are there two similar formats? Can this be added as part of ncnn support instead of copying code? Tencent/tnn is not available.

nihui:
Hi

As far as I know, tnn is yet another deep-learning inference framework that will be open on June 10th, and the tnn model format has referenced ncnn's and made some of its own extensions.
But it is a new format which ncnn library cannot load, while tnn will support ncnn model format directly and its own format at the same time.

I am not sure if it is a good idea to add these new things as part of ncnn support. It may be hard to maintain format and operator compatibility as both projects are evolving and progressing in its own.

masterwan96:
Hi~ 
TNN is a brand-new neural network architecture of Tencent. It is similar to ncnn but is not the same architecture, so it cannot be merged into one project. I attached the model of tnn in the attachment, hope to get your opinion




从网易126邮箱发来的云附件
fcn_resnet101.onnx.opt.onnx.tnnproto (11.26K, 2020年6月23日 19:01 到期)
下载
fcn_resnet101.onnx.opt.onnx.tnnmodel (207M, 2020年6月23日 19:01 到期)
下载

masterwan96:
https://drive.google.com/drive/folders/1iszQIHIximiPMwYAfjZ0kaA0-a-LNPNa?usp=sharing
There are some models of tnn inside, you can check it~ 
Thank you very much.

masterwan96:
TNN is open source up to now~
https://github.com/Tencent/TNN.git

lutzroeder:
`Unknown node type 'undefined' in 'tnn'.` when opening `version-slim-320_simplified.tnnmodel`.

masterwan96:
I need version-slim-320_simplified.proto to find out  where is wrong.
Proto file and model file are necessary for tnn

lutzroeder:
It is not clear from the description why both ncnn and tnn formats exist. Why was there another format invented that is almost identical to an existing format? There are 30+ other formats for storing deep learning models? Could you modify TNN to support an existing format?

lutzroeder:
`npx eslint .` fails with `17 problems (17 errors, 0 warnings)`.

masterwan96:
TNN and NCNN are developed independently. The TNN model format has been maintained within Tencent for many years, and many business departments within Tencent use the TNN model format for business deployment. If the model format is changed,  Businesses previously deployed using the TNN model format will not run

lutzroeder:
Maybe they shouldn't be developed independently as this creates unnecessary code duplication and maintenance cost? This sounds overly complicated and maybe the right approach is to remove both tnn and ncnn support as those formats seem to be unstable and constantly changing? Have you considered adopting a more standard format instead of inventing and duplicating formats? Businesses can be migrated to new formats to reduce the number of overall formats.

nihui:
ncnn is a fully community-driven open source project, with 100+ contributors.

tnn used to be the commercial framework for tencent's interior, and at the beginning the model format was exactly the same as ncnn, but a lot of modifications were made later.
These modifications may be designed to prevent business models from being stolen.
For strict confidentiality, ncnn's owner (me) cannot view the tnn code or participate in the development of tnn until recent months.

As I mentioned earlier, tnn actually supports the ncnn model format and will continue to support it in the future. Yeah, there are businesses being migrating to ncnn format.

So the common standard format is already there.

This PR is intended to make it easier to view some of the existing tnn models.

neiltian-tencent:
Hi, lutzroeder。I'm the admin of TNN 。TNN can only read and parse some ncnn models but not all。At the model level, TNN  is different from ncnn，especially for operator fusion and low precision quantization，TNN will develop independently。In the future, TNN will invest a lot of manpower to maintain and support the construction of open source communities。
In the two days since open source, the TNN star has been close to 1,000。
As the standard of visualization tools, netron is very important for TNN users. 
Happy and expect to see netron can support TNN model structure。

lutzroeder:
@neiltian-tencent, by that line of reasoning, Netron has lots of stars, maybe TNN should switch to a standard format that is supported by Netron instead of inventing yet another format that is similar to all the other formats that already exist? In fact that would make it easier for all TNN users not having to use a cryptic format that is different from all other formats and offers no benefit. The code in this PR is a copy of the ncnn code with mostly everything added making it harder to maintain or needs to be rewritten.

neiltian-tencent:
@lutzroeder thanks for your suggestion. For the code problem in PR, we will continue to solve it. We will also evaluate the feasibility of adopting the ncnn model completely.

lutzroeder:
#521

lutzroeder:
@neiltian-tencent the model files are not accessible: 

```
~/TNN/model: ./download_model.sh
download mobilenet_v2.onnx.opt.onnx.tnnmodel failed
download shufflenet_v2.onnx.opt.onnx.tnnmodel failed
```

lutzroeder:
This code is used to detect the signature but [these](https://github.com/Tencent/TNN/tree/master/benchmark/benchmark-model) files are missing the signature: 

```
if (identifier.endsWith('.tnnproto') || identifier.endsWith('.rapidproto')) {
    let text = context.text;
    const lines = text.split(/\r?\n/);
    const header = lines.shift().split(' ').slice(3,-1);
    const signature = header.shift();
    if (signature === '4206624770') {
        return true;
    }
```

lutzroeder:
There are no `.tnnproto.tnnmodel` files available.

neiltian-tencent:
@lutzroeder thanks for your refactoring. During the process from closed source to open source, we have made some changes and we will fix these problems immediately.

PR #526, opened 2020-06-22
Fix 'Div' and 'Sub' OP

masterwan96:
Fixed the bug that these two layers cannot display weight

### DIFF
--- a/src/tnn.js
+++ b/src/tnn.js
@@ -780,6 +780,8 @@ tnn.LayerResourceReader = class {
                         resource.slope = raw(reader);
                         break;
                     }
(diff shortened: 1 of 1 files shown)


lutzroeder:
Can you share sample files?

masterwan96:
This problem will be encountered when the Sub or Div layer has weight, and cannot open .tnnmodel.
It shows  'Unknown layer resource type 'Sub' in 'xxx.tnnmodel' in 'xxx.tnnmodel'. ' 
error
error2

lutzroeder:
Can you share a sample model file to reproduce the issue?

Issue #497, opened 2020-05-22
PyTorch: Loading 10 GB model fails


psyhtest:
Loading the [~10 GB DLRM model](https://dlrm.s3-us-west-1.amazonaws.com/models/tb0875_10M.pt) from [MLPerf Inference](https://github.com/mlperf/inference/tree/db0b7eb368f009adf703d073765aa212d1f3a026/v0.5/recommendation) [failed](https://github.com/mlperf/inference/issues/608#issue-623152831) on an Ubuntu 18.04 machine with 32 GB RAM:
- from the browser: `File read error '0'.`
- from the command line: `Error loading model. Unsupported file content for extension '.pt' in 'tb0875_10M.pt'.` (memory usage shoot up to 27 GB)

I would completely understand if the same thing happened with the ~100 GB model :), but maybe the ~10 GB one could be supported still?

P.S. Netron is a beautiful and useful tool - thank you so much for it @lutzroeder!

lutzroeder:
[ColorizeVideo_PretrainOnly_gen.pth](https://www.dropbox.com/s/avzixh1ujf86e8x/ColorizeVideo_PretrainOnly_gen.pth?dl=0)

lutzroeder:
https://console.cloud.google.com/storage/browser/vit_models/imagenet21k

lutzroeder:
The desktop app now supports this. The Web and Python versions are constrained by the 2 GB browser limit.

PR #584, opened 2020-08-25
CPU/Memory improvement: Replaced Array.concat where possible with Array.push(...) which is way faster

NikolasMarkou:
Performance and memory improvement: where possible I replaced the Array.concat with Array.push.
According to [this](https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki) Array.push is way faster.
```
The improvement has a lot to do with the fact that Array.concat creates a new array while Array.push modifies the first array. The additional work Array.concat does to add the elements from the first array to the returned array is the main reason for the slowdown.
```
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,6 +115,7 @@ setuptools.setup(
             'openvino.js', 'openvino-metadata.json', 'openvino-parser.js',
             'paddle.js', 'paddle-metadata.json', 'paddle-proto.js',
             'pytorch.js', 'pytorch-metadata.json', 'python.js',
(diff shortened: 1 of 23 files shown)


NikolasMarkou:
#497 this will allow loading bigger models (not sure if 10gb), concat spikes are avoided

lutzroeder:
Is there a specific scenario where this is noticeable? 

Most of these calls operate on very small local arrays. For really large arrays this might be a problem:

```
var array1 = []
var array2 = new Array(400000);
array1.push(...array2);

RangeError: Maximum call stack size exceeded
```

PR #598, opened 2020-09-18
fix bug: multi-input/output for Tengine

pierricklee:
Hi Lutz,

When rendering multi-input or multi-output Tengine models, Netron failed like this : 
image

While the real model is like :
image

This is because in subgraph decoding, it gets `subgraph.inputs` are actually nodes but not tensors (also works for `subgraph.outputs`). 

I modified and successfully render like this :
image

ChangeLog:
1. Allow `Const` operators be pushed into subgraph ;
2. Read real `graph inputs` and `graph outputs` by `node id` ; 
3. Do not render `Const` operators

Here is the `tmfile` model attached so you can test.
[test_multi_input_output.zip](https://github.com/lutzroeder/netron/files/5244386/test_multi_input_output.zip)

----------------
Thanks,

Pierrick

### DIFF
--- a/source/tengine.js
+++ b/source/tengine.js
@@ -79,7 +79,7 @@ tengine.Graph = class {
         }
 
         for (const node of graph.nodes) {
-            if (node.type !== 'INPUT') {
+            if (node.type !== 'INPUT' && node.type !== 'Const') {
                 this._nodes.push(new tengine.Node(metadata, node, tensors));
             }
         }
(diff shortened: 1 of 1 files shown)


pierricklee:
Sorry I forgot the source onnx model. Please use these two models.
[test_multi_input_output.zip](https://github.com/lutzroeder/netron/files/5244410/test_multi_input_output.zip)

And also, you could create a test model by yourself and convert it to `tmfile` by [convert_tool](https://convertmodel.com/).

PR #614, opened 2020-10-13
Acuity support

hawk081:
Hi,

This patch enables the support of Acuity models, please help to review the patch, any suggestions are welcome.

Acuity is an AI framework that can convert the models of other framework(such as tensorflow, tflite, caffe, onnx, etc) to Acuity models, then generate high performance code for embedded platforms.

`.json` file is the Acuity model file which contains the model structure, and `.data` file will be loaded for weights and biases, `.quantize` file is for quanzation parameters. the output tensors' shape of each layer will be inferred if the input shape(s) are provied in model files.

Some popular models converted by Acuity can be found here: https://verisilicon.github.io/acuity-models/

Thanks.

#657

### DIFF
--- a/test/models.json
+++ b/test/models.json
@@ -1,5 +1,11 @@
 [
   {
+    "type":   "acuity",
+    "target": "inception_v1.json",
+    "source": "https://raw.githubusercontent.com/VeriSilicon/acuity-models/master/models/inception_v1/inception_v1.json",
+    "link":   "https://github.com/VeriSilicon/acuity-models"
+  },
+  {
     "type":   "any",
     "target": "config.json",
     "source": "https://github.com/lutzroeder/netron/files/5398758/config.json.zip[config.json]",
--- a/source/acuity-metadata.json
+++ b/source/acuity-metadata.json
@@ -1,0 +1,3067 @@
+[
+  {
+    "name": "a_times_b_plus_c",
+    "schema": {
+      "category": "Layer",
+      "inputs": [
+        {"name": "in0", "description": "in"},
+        {"name": "in1", "description": "in"},
+        {"name": "in2", "description": "in"}
+      ],
+      "constants": [],
+      "outputs": [
+        {"name": "out0", "description": "out"}
+      ],
+      "attributes": []
+    }
+  },
+  {
+    "name": "abs",
+    "schema": {
+      "category": "Layer",
+      "inputs": [
+        {"name": "in0", "description": "in"}
+      ],
+      "constants": [],
+      "outputs": [
+        {"name": "out0", "description": "out"}
+      ],
+      "attributes": []
+    }
+  },
+  {
+    "name": "add",
+    "schema": {
+      "category": "Layer",
+      "inputs": [
+        {"name": "in0", "description": "in"},
+        {"name": "in1", "description": "in"}
+      ],
+      "constants": [],
+      "outputs": [
+        {"name": "out0", "description": "out"}
+      ],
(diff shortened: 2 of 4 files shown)


hawk081:
Hi Lutz,

Is there anything need to be modified ?

Thanks.

lutzroeder (changes requested):
1. There are multiple similar formats for deep learning models already. Can Acuity use an existing format instead of creating another duplicate format?
2. `jsyaml` should not be checked in as obfuscated code. For JSON, Protocol Buffers and FlatBuffers are custom implementation using `base.TextDecoder` is checked in. A similar approach should be followed for YAML. Note that using JSON and YAML and Pickle (3 different formats) for a single model is a fragile design. Ideally there should be no YAML dependency.

lutzroeder:
- [ ] There are dozens of similar formats for deep learning models already. Can Acuity use an existing format instead of investing yet another one?
- [ ] `jsyaml` should not be checked in as obfuscated code. For JSON, Protocol Buffers and FlatBuffers there are custom implementations for Netron and a similar approach would be needed for YAML. Using JSON and YAML and binary (3 different formats) for a single model is a fragile design. Ideally there should be no YAML dependency.

test/models.json
```diff
@@ -5776,5 +5776,11 @@
     "target": "j48model.model",
     "source": "https://raw.githubusercontent.com/PTaati/wekaTree2python/master/j48model.model",
     "error":  "Unsupported type 'weka.classifiers.trees.J48' in 'j48model.model'."
+  },
+  {
+    "type":   "acuity",
+    "target": "inception_v1.json",
+    "source": "https://raw.githubusercontent.com/VeriSilicon/acuity-models/master/models/inception_v1/inception_v1.json",
```
lutzroeder, line 5783: The file is sorted alphabetically by `type`. This is only testing the JSON code, not the YAML and binary readers. Please add a more comprehensive sample.

lutzroeder (changes requested):
- [ ] There are dozens of similar formats for deep learning models already. Can Acuity use an existing format instead of investing yet another one?
- [ ] `jsyaml` should not be checked in as obfuscated code. For JSON, Protocol Buffers and FlatBuffers there are custom implementations for Netron and a similar approach would be needed for YAML. Using JSON and YAML and binary (3 different formats) for a single model is a fragile design. Ideally there should be no YAML dependency.

hawk081:
Thanks!

1. Current format of `.json`, `.data` and `.quantize` are used in our production, It is difficult to migrate to other formats.
2. We will follow the `JSON` to implement a YAML parser.
3. For some historical reasons, we have to use these 3 files to describe the model for now.

lutzroeder:
> Current format of .json, .data and .quantize are used in our production, It is difficult to migrate to other formats.

It will only get more difficult over time. This is an overcomplicated design with lots of moving parts. Better to fix it early one while you still can. Ideally you should consider what value the framework actually provides and if it does need yet another new file format or can just use an existing one.

hawk081:
Thanks for your suggestions.

I will remove the code of loading .quantize and .data, let netron just handle the model structure.

lutzroeder:
Why is this format needed? There are a dozen other deep learning formats that are more or less identical.

lutzroeder:
Why is this format needed? There are a dozen other deep learning formats that are more or less identical.

hawk081:
Thanks! When we started the project at 2016, there weren't suitable formats for us. so ...

PR #641, opened 2020-12-01
supplement tensor name for MindSpore Lite

YvetteLaw:
For MindSpore Lite:
- supplement tensor name
- supplement 'string' dataType

### DIFF
--- a/source/mslite.js
+++ b/source/mslite.js
@@ -67,7 +67,8 @@ mslite.Graph = class {
             const data = tensor.data;
             const type = new mslite.TensorType(tensor.dataType, tensor.dims);
             const initializer = (data && data.length > 0) ? new mslite.Tensor(type, tensor.data) : null;
-            return new mslite.Argument(index.toString(), tensor, initializer);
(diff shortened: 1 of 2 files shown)


source/mslite.js
```diff
@@ -67,7 +67,7 @@ mslite.Graph = class {
             const data = tensor.data;
             const type = new mslite.TensorType(tensor.dataType, tensor.dims);
             const initializer = (data && data.length > 0) ? new mslite.Tensor(type, tensor.data) : null;
-            return new mslite.Argument(index.toString(), tensor, initializer);
+            return new mslite.Argument(tensor.name, tensor, initializer);
```
lutzroeder, line 70: This will fail for existing models as `tensor.name` might be null?

lutzroeder (changes requested):

source/mslite.js
YvetteLaw: The tensor name will be index.toString() if it is not defined. I also updated the test models.

PR #658, opened 2020-12-26
modify default tensorType of MSLite

YvetteLaw:
fix a bug

### DIFF
--- a/source/mslite.js
+++ b/source/mslite.js
@@ -436,7 +436,35 @@ mslite.TensorType = class {
     constructor(dataType, dimensions) {
         switch (dataType) {
             case 0:  this._dataType = "?"; break;
+            case 1:  this._dataType = "MetaTypeType"; break;
+            case 2:  this._dataType = "MetaTypeAnything"; break;
+            case 3:  this._dataType = "MetaTypeObject"; break;
+            case 4:  this._dataType = "MetaTypeTypeType"; break;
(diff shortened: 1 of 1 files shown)


lutzroeder:
What exactly is the bug and is there a repro? The code should throw for an unknown data type and the specific issue should get fixed instead of silently catching all failures.

YvetteLaw:
> What exactly is the bug and is there a repro? The code should throw for an unknown data type and the specific issue should get fixed instead of silently catching all failures.

https://github.com/mindspore-ai/mindspore/blob/master/mindspore/core/ir/dtype/type_id.h

lutzroeder:
The right approach is to show a useful name and fail if an unknown or new one is added. If the value is saved to a persistent model it has some meaning and users might want to know about that instead of seeing a `?`. Some of these do not seem to be data types but they are stored as data types -- is that a design that should be fixed? Is there a sample model file to reproduce the issue you are trying to fix?

YvetteLaw:
> The right approach is to show a useful name and fail if an unknown or new one is added. If the value is saved to a persistent model it has some meaning and users might want to know about that instead of seeing a `?`. Some of these do not seem to be data types but they are stored as data types -- is that a design that should be fixed? Is there a sample model file to reproduce the issue for this fix?

I've modified them with their name. 

"Some of these do not seem to be data types but they are stored as data types -- is that a design that should be fixed? "-----It's our design.

PR #671, opened 2021-01-22
add subGraph function to MSLite

YvetteLaw:
We have added the subGraph function to MSLite. Please merge this PR.

### DIFF
--- a/source/mslite.js
+++ b/source/mslite.js
@@ -44,7 +44,16 @@ mslite.Model = class {
             const version = this._format.substring(format.length).replace(/^v/, '');
             this._format = format + 'v' + version;
         }
-        this._graphs = [ new mslite.Graph(metadata, model, model) ];
+        this._graphs = [];
+        const subgraphs = model.subGraph;
+        if (subgraphs === []) {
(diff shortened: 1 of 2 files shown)


lutzroeder:
@YvetteLaw can you add a sample model in `./test/models.json`?

YvetteLaw:
> @YvetteLaw can you add a sample model in `./test/models.json`?

Please see the model 'control_flow_ut_while_0122.ms'

lutzroeder:
This PR seems to break compatibility with the 1.0 format. Also, please add back a 1.0 test file for regression testing. It is not acceptable that the format keeps changing and users will have a bad experience when loading existing models. If this cannot be resolved maybe MSLite support should be removed as the format is not stable.

YvetteLaw:
> This PR seems to break compatibility with the 1.0 format? Please add back a 1.0 test file for regression testing. It is not acceptable that the format keeps changing and users will have a bad experience when loading existing models. If this cannot be resolved maybe MSLite support should be removed as the format is not stable.

I've added test models with 1.0.0 format

lutzroeder:
`subgraph === []` does not work .

PR #700, opened 2021-03-15
Uupdate mslite to r1.2

YvetteLaw:
Thanks for your support of our MindSpore Lite r1.2. Files are our update to r1.2.

### DIFF
--- a/source/mslite-metadata.json
+++ b/source/mslite-metadata.json
@@ -1,2876 +1,3598 @@
 [
   {
     "name": "Abs",
-    "inputs": [
-      { "name": "x" }
-    ],
-    "outputs": [
-      { "name": "y" }
-    ]
-  },
-  {
-    "name": "AbsGrad"
+    "schema": {
+      "inputs": [
+        {
+          "name": "input_x"
+        }
+      ],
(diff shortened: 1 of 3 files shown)


YvetteLaw:
I modify tools/flatc.js because I think the contents of the "include" file should be placed in front of the contents of this file when generating a *-schema.js. Otherwise, there may be a problem that the reference could not be found.

lutzroeder:
> Otherwise, there may be a problem that the reference could not be found.

@YvetteLaw, can you provide steps to reproduce this issue?

YvetteLaw:
I modify tools/flatc.js because I think the contents of the "include" file should be placed in front of the contents of this file when generating a *-schema.js. Otherwise, there may be a problem that the reference could not be found.

> > Otherwise, there may be a problem that the reference could not be found.
> 
> @YvetteLaw, can you provide steps to reproduce this issue?

I found this issue in January. When I run './tools/mslite schema', flatc.js said that It could not find the referenced content, so I modified it as shown in this PR. I tried to reproduce this issue but failed. I think you may have fixed this issue during this period, or that this may be a problem of our code itself at that time.

PR #701, opened 2021-03-16
update tengine: add OneFlow, Horizon and Bitman model support; add op…

pierricklee:
add OneFlow, Horizon, and Bitman model support;
add tengine operators.
fix Reshape display.

### DIFF
--- a/source/tengine-metadata.json
+++ b/source/tengine-metadata.json
@@ -219,6 +219,21 @@
     ]
   },
   {
+    "name": "Reshape",
+    "schema": {
+      "category": "Shape",
+      "attributes": [
+        { "name": "is_mxnet", "type": "int32", "default": 0 },
+        { "name": "reverse", "type": "int32", "default": 0 },
+        { "name": "shape", "type": "int32[]", "default": [] }
+      ],
+      "inputs": [
+        { "name": "input" },
+        { "name": "shape" }
+      ]
+    }
+  },
+  {
     "name": "DropOut",
     "category": "Dropout"
   },
@@ -867,6 +882,16 @@
     ]
   },
   {
+    "name": "Scatter",
+    "schema": {
+      "category": "Layer",
+      "attributes": [
+        { "name": "axis", "type": "int32", "default": 0 },
+        { "name": "is_onnx", "type": "boolean", "default": false }
+      ]
+    }
+  },
+  {
     "name": "Unsqueeze",
     "category": "Transform",
     "attributes": [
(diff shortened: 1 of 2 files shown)


source/tengine.js
```diff
@@ -637,6 +642,7 @@ tengine.ModelFileReader = class {
         this._graphs = [];
         const subgraphOffsets = reader.uint32s();
         for (const subgraphOffset of subgraphOffsets) {
+            subgraphOffsets.length =1 ;
```
lutzroeder, line 645: Why is this required?

source/tengine.js
```diff
@@ -548,8 +548,8 @@ tengine.ModelFileReader = class {
         register(20, 0, 'ReLU', [ 'f' ]);
         register(21, 0, 'ReLU6', []);
         register(22, 0, 'Reorg', [ 'i' ]);
-        register(23, 0, 'Reshape', [ 'i', 'i', 'i', 'i', 'i', 'i' ]);
-        register(23, 2, 'Reshape', [ 'i', 'i', 'i[]' ]);
+        // register(23, 0, 'Reshape', [ 'i', 'i', 'i', 'i', 'i', 'i' ]);
```
lutzroeder, line 551: Test regression. Run `node ./test/models.js tengine`.

source/tengine-metadata.json
```diff
@@ -218,6 +218,21 @@
       { "name": "detect_num" }
     ]
   },
+  {
+    "name": "Reshape",
```
lutzroeder, line 222: Duplicate.

source/tengine-metadata.json
```diff
@@ -866,6 +881,16 @@
       { "name": "type", "type": "int32", "default": 0 }
     ]
   },
+  {
+    "name": "Scatter",
```
lutzroeder, line 885: Sort alphabetically.

lutzroeder (changes requested):

source/tengine.js
pierricklee: Hi Lutz, 

This is for Tengine's subgraph strategy. Tengine sets the main subgraph as subgraph[0], other subgraphs[1-n] are cut from subgraph[0]. These subgraphs may have different data type, due to a hybrid-precision strategy. For example, subgraph[1] is float32 while subgraph[2] uses uint8, they are both sub-graph to subgraph[0]. 

So basically, the subgraph[0] is not actually a 'subgraph'. It is 'the graph'. 
Subgraphs shouldn't be shown when rendering, only 'the graph' will be shown.

[hybrid-precision subgraph  model.zip](https://github.com/lutzroeder/netron/files/6471719/hybrid-precision.subgraph.model.zip)

We can use this model to test. It's also an experimental feature in Tengine.

PR #709, opened 2021-03-24
Set bypass cache query parameter on model URL in key=value format

### LINKED ISSUE #708: Appending of timestamp to model URL breaks mlflow artifact API
- Netron app and version: 4.8.1
- mlflow version: 1.7.1

As described in #697 I want to automatically open netron with artifacts saved in mlflow. I got it working now, but stumbled over a probelm. In https://github.com/lutzroeder/netron/blob/main/source/index.js#L438 the current time is appended to the model URL as a query parameter. However, since it only contains a value and not a key, it somehow breaks the mlflow URL parser, resulting in an internal server error. My workaround is to simply remove this line in my deployment, but I would like to understand what it is for and whether it could be changed to a key value format. 

Unfortunately, it is a bit difficult to publicly reproduce the error, as one would need a public mlflow instance and enable CORS for netron there. I hope the problem is stll clear enough.

Steps to Reproduce:

1. Track a keras model as .h5 file as an artifact in mlflow. 
2. Try to open the url in netron, an example URL would look like http://netron.app/?identifier=model.h5&url=http://mlflow.my-domain.com/get-artifact?path%3Dmodel.h5%26run_id%3D79a796f0dd5449b58e6e7e6d0a638f51
3. An internal server error is triggered in mlflow.

JarnoRFB:
Fixes #708. The value only parameter otherwise breaks some query parameter parsers,
such as the one of mlflow.

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -435,6 +435,7 @@ host.BrowserHost = class {
     }
 
     _openModel(url, identifier) {
+        url = url + ((/\?/).test(url) ? "&" : "?") + "netronBypassCache=" + (new Date()).getTime();
         this._view.show('welcome spinner');
         this._request(url).then((buffer) => {
             const context = new host.BrowserHost.BrowserContext(this, url, identifier, buffer);
--- a/source/index.js
+++ b/source/index.js
@@ -435,7 +435,6 @@ host.BrowserHost = class {
     }
 
     _openModel(url, identifier) {
-        url = url + ((/\?/).test(url) ? "&" : "?") + (new Date()).getTime();
         this._view.show('welcome spinner');
         this._request(url).then((buffer) => {
             const context = new host.BrowserHost.BrowserContext(this, url, identifier, buffer);


JarnoRFB:
One could of course argue that this behavior is the fault of the query parameter parser, but if the extra parameter can be omitted it might be easier to fix it here.

lutzroeder:
@JarnoRFB See [Bypassing the cache](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#bypassing_the_cache). Fix either the query parameter parser or by using a supported pattern.

source/index.js
```diff
@@ -435,7 +435,6 @@ host.BrowserHost = class {
     }
 
     _openModel(url, identifier) {
-        url = url + ((/\?/).test(url) ? "&" : "?") + (new Date()).getTime();
```
lutzroeder, line 438: @JarnoRFB See [Bypassing the cache](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#bypassing_the_cache). Fix either the query parameter parser or by using a supported pattern.

lutzroeder (changes requested):

JarnoRFB:
@lutzroeder Thanks for enlightening me on this topic! I now set the timestamp in key=value format, which works equally well. Let me know what you think of it.

PR #715, opened 2021-04-03
OpenVINO Add precision type to inputs and int4 support

Vladimir9812:
Hi, @lutzroeder, in this PR has been added support for int4 and uint4 layer precisions and have been enabled precisions for layer inputs

### DIFF
--- a/source/openvino.js
+++ b/source/openvino.js
@@ -125,16 +125,18 @@ openvino.Graph = class {
         this._outputs = [];
         this._arguments = {};
 
+        const layers = new Map(net.layers.map((entry) => [ entry.id, entry ]));
+
         for (const layer of this._const(net.layers, net.edges)) {
             const inputs = layer.inputs.map((input) => {
                 const to = layer.id + ':' + input.id;
-                const [outputLayerId, outputId] = net.edges[to].split(':');
-                const outputLayer = net.layers.find((layer) => layer.id === outputLayerId);
-                if (outputLayer) {
+                const [outputLayerId, outputId] = net.edges[to] ? net.edges[to].split(':') : [];
+                const outputLayer = layers.get(outputLayerId);
+                if (outputLayer && outputId) {
                     const output = outputLayer.outputs.find((output) => output.id === outputId);
                     input.precision = output.precision;
                 }
-                return this._argument(layer.id, layer.precision, input, net.edges);
+                return this._argument(layer.id, input.precision || layer.precision, input, net.edges);
             });
             const outputs = layer.outputs.map((output) => this._argument(layer.id, output.precision || layer.precision, output, null));
             switch (layer.type) {
@@ -959,7 +961,9 @@ openvino.TensorType = class {
             case 'bool':    this._dataType = 'boolean'; break;
             case 'boolean': this._dataType = 'boolean'; break;
             case 'bin':     this._dataType = 'bit'; break;
-            default:        this._dataType = '?'; break;
+            case '':        this._dataType = '?'; break;
+            case null:      this._dataType = '?'; break;
+            default:        throw new openvino.Error("Unknown precision '" + JSON.stringify(precision) + "'.");
         }
         this._shape = shape;
     }
--- a/source/openvino.js
+++ b/source/openvino.js
@@ -822,9 +822,26 @@ openvino.Tensor = class {
             return context;
         }
 
-        context.index = 0;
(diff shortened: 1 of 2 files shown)


source/openvino.js
```diff
@@ -126,7 +126,16 @@ openvino.Graph = class {
         this._arguments = {};
 
         for (const layer of this._const(net.layers, net.edges)) {
-            const inputs = layer.inputs.map((input) => this._argument(layer.id, layer.precision, input, net.edges));
+            const inputs = layer.inputs.map((input) => {
+                const to = layer.id + ':' + input.id;
+                const [outputLayerId, outputId] = net.edges[to].split(':');
+                const outputLayer = net.layers.find((layer) => layer.id === outputLayerId);
```
lutzroeder, line 132: Lookup map and `find` can be inefficient.

source/openvino.js
```diff
@@ -126,7 +126,16 @@ openvino.Graph = class {
...
+                if (outputLayer) {
+                    const output = outputLayer.outputs.find((output) => output.id === outputId);
+                    input.precision = output.precision;
```
lutzroeder, line 135: * What exactly is this code doing?
* Can the `precision` of the output always assumed to be the same as the input.
* `input` contains source data that should be copied instead of modified.

lutzroeder (changes requested):

source/openvino.js
Vladimir9812: What lookup map can you suggest?

source/openvino.js
Vladimir9812: Initially in XML file inputs don't have precision property, unlike outputs. So, I try to match inputs and outputs according to the edges map and assign precisions to inputs.
Yes, precision to the output always assumed to be the same as the input.

source/openvino.js
lutzroeder: `const layers = new Map(net.layers.map((entry) => [ entry.id, entry ]))` and `layers.get(outputLayerId))`.

source/openvino.js
```diff
@@ -126,7 +126,16 @@ openvino.Graph = class {
         this._arguments = {};
 
         for (const layer of this._const(net.layers, net.edges)) {
-            const inputs = layer.inputs.map((input) => this._argument(layer.id, layer.precision, input, net.edges));
+            const inputs = layer.inputs.map((input) => {
+                const to = layer.id + ':' + input.id;
```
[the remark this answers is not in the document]
Vladimir9812: Thank you, @lutzroeder, I will fix all your comments tomorrow during the day

source/openvino.js
```diff
@@ -936,21 +945,21 @@ openvino.TensorType = class {
             case 'f32':     this._dataType = 'float32'; break;
             case 'fp32':    this._dataType = 'float32'; break;
             case 'bf16':    this._dataType = 'bfloat16'; break;
+            case 'i4':      this._dataType = 'int4'; break;
```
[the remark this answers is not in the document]
Vladimir9812: This is a model with int4 for test
[int4-model.zip](https://github.com/lutzroeder/netron/files/6259225/int4-model.zip)

Vladimir9812:
@lutzroeder, I don't have any information about correctly decoding int4 dataType. So I've added a stub so far, but it is a temporary decision. I corrected all the other comments and PR is waiting for a second review round

source/openvino.js
lutzroeder: @Vladimir9812 "assumed" to be the same as in this code makes this assumption but it could be wrong or is this always the case for all OpenVINO models?

source/openvino.js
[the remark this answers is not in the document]
lutzroeder: Can you provide a smaller model? 10 MB is a lot of a test file to test a data type.

source/openvino.js
Vladimir9812: @lutzroeder, sorry for my long answers...Yes, it is true for all OpenVINO models.
Thank you for merge. Can I find a smaller int4 model and attach it to this PR?

lutzroeder: Yes, that would be great.

Vladimir9812: [simple_model_int4.zip](https://github.com/lutzroeder/netron/files/6278507/simple_model_int4.zip)

PR #735, opened 2021-04-28
Add prefix option for serving behind proxy

SpontaneousDuck:
I do most of my work on a remote machine using JupyterHub which is a proxy in front of my Jupyter instance. I use [jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy) to expose ports through the proxy (ex. http://192.168.1.2/notebook/mine/temp_nb/proxy/8081/ where netron is serving on port 8081). 

The current version of netron does not work with this setup since it serves the model file from the top level of the server path. I added a parameter to allow the user to tell the server what path it is being served on underneath the top address and prepend that to the serving path of the model. This should not have any affect for users who are using netron normally.

Not sure if you want this in the main repository but figured I would share the work I did in case anyone else has the same use case. Thanks for the great work!

### DIFF
--- a/source/server.py
+++ b/source/server.py
@@ -53,7 +53,7 @@ class HTTPRequestHandler(BaseHTTPRequestHandler):
         buffer = None
         data = '/data/'
         if status_code == 0:
-            if pathname[-1] == '/':
+            if pathname == '/' or pathname == self.prefix:
                 meta = []
                 meta.append('<meta name="type" content="Python">')
                 meta.append('<meta name="version" content="' + __version__ + '">')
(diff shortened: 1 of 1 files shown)


lutzroeder:
The other files like `index.js` would still be served from the top level folder? Can you provide some context why this is an issue for the model file and not any of the other files?

SpontaneousDuck:
Sure! So I use JupyterHub which puts a proxy in front of your notebook and serves it under the main server IP. Ports in each Jupyter instance are also served under a subpage. For Netron, this would look like:  http://192.168.1.2/notebook/mine/temp_nb/proxy/8081/ . All the static files are still served properly from this route (ex. http://192.168.1.2/notebook/mine/temp_nb/proxy/8081/index.js). Even the model is served at the correct route (http://192.168.1.2/notebook/mine/temp_nb/proxy/8081/data/model.onnx).

What does not work is the insertion of the route to the mdoel file in the `meta` tag by the `HTTPRequestHandler` in `server.py`. This tells the web app that the model is always served at `<ip>/data/model.oxx`. What this PR does is let you specify the path under the server IP where netron is running. This allows the `HTTPRequestHandler` to correctly populate the `<meta>` tag. 

Sorry if that is a little wordy!

lutzroeder:
If other files work this seems to be an issue with the client prepending the location. Would 7bb852b resolve this?

PR #736, opened 2021-04-29
Add ncnn layer activation/op type display support.

zhiliu6:
Add support for InnerProduct, BinaryOp, Eltwise, Pooling, Convolution, Deconvolution type support.

### DIFF
--- a/source/ncnn.js
+++ b/source/ncnn.js
@@ -207,6 +207,7 @@ ncnn.Node = class {
         this._inputs = [];
         this._outputs = [];
         this._attributes = [];
+        this._chain = [];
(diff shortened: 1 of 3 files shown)


source/ncnn.js
```diff
@@ -287,6 +294,27 @@ ncnn.Node = class {
                 this._weight(blobReader, 'bias', [ bias_data_size ], 'float32');
                 break;
             }
+            case 'BinaryOp': {
```
lutzroeder, line 297: Is there a way to improve this that does not require changing `type` semantics? For example, always show activation type even if "Show Attributes" is disabled?

Issue #784, opened 2021-08-13
TensorFlow: signature support


TomWildenhain-Microsoft:
- Netron app and version: 5.1.0
- OS and browser version: Windows 10 desktop

I love the new subgraph UI to access subgraphs via nodes, but the removal of a global subgraph list for TF saved models makes most subgraphs inaccessible.  (only the save/restore subgraphs are referenced by the metagraph. Inference graphs are referenced by signatures but not directly called by the metagraph).

Example model:
https://tfhub.dev/google/tf2-preview/nnlm-en-dim50/1?tf-hub-format=compressed

lutzroeder:
@TomWildenhain-Microsoft which subgraph in the sample file is not accessible and how is it accessed by signatures?

TomWildenhain-Microsoft:
The primary graph of interest is __inference___call___110 which calls __inference__preprocess_88 (the later can also be called independently)

![image](https://user-images.githubusercontent.com/67606533/129610708-dee018ce-daae-490c-8e5f-8d21958cb26b.png)

The signatures can be found using the saved model cli with `saved_model_cli show --dir C:\Users\tomwi\Downloads\nnlm --all` and are:

```
Defined Functions:
  Function Name: '__call__'
    Option #1
      Callable with:
        Argument #1
          sentences: TensorSpec(shape=(None,), dtype=tf.string, name='sentences')

  Function Name: '_preprocess'
    Option #1
      Callable with:
        Argument #1
          sentences: TensorSpec(shape=(None,), dtype=tf.string, name='sentences')
```

This model is of course just one example (simplest I could find).

Fannettembrown:
TF1 is a hub format, you also can use segments model Url as well.

lutzroeder:
```
tf/half_plus_two_saved_model.pb
tf/saved_model_xxx.pb

tf/0342.zip
tf/resnet_v2_fp16_savedmodel_NHWC_saved_model.pb

tf/tensorflow_issue_9169_saved_model.pbtxt
tf/myTag.pb
```

PR #785, opened 2021-08-16
Re-add function list in graph sidebar for TF

TomWildenhain-Microsoft:


lutzroeder:
See #784. The right way to resolve this would be supporting signatures?

TomWildenhain-Microsoft:
Supporting signatures would be nice, but I think random access to subgraphs by name is useful regardless.  Keep in mind that TF isn't a recursive graph structure; it is set of function names mapped to graphs, which can call each other.  

Graphs for initializers would be inaccessible even if signatures were implemented.

lutzroeder:
Yes, this requires some exploring what the right design should be. Mixing functions and graphs into a dropdown is likely not the right design.

TomWildenhain-Microsoft:
No problem.  I'll just stick to an old version of Netron for now when viewing saved models.

PR #787, opened 2021-08-20
Add shape inference for acuity operations

hawk081:
Add shape inference for more acuity operations


lutzroeder:
```
~/netron_acuity: node ./test/models.js acuity
acuity/inception_v1.json
parameters.shape.map is not a function in 'inception_v1.json'.
```

lutzroeder (changes requested):
```
~/netron_acuity: node ./test/models.js acuity
acuity/inception_v1.json
parameters.shape.map is not a function in 'inception_v1.json'.
```

hawk081:
The parameters.shape of `reshape` is string in old .json, I support it in new commit.

PR #795, opened 2021-08-30
xmodel: visualization

xy-xlnx:
Hi, lutzroeder:
We really appreciate your work about Netron, which is widely used by us. I tried to optimize the visualization of xmodel. Please have a look at this patch, could these changes be merged into your main branch ?

New Features:
1. Visualize the subgraphs for the .xmodel files.
2. Display the attributes of tensors.
3. Change the color.

Some Examples:
image
image

#718

### DIFF
--- a/source/om.js
+++ b/source/om.js
@@ -437,52 +437,12 @@ om.TensorShape = class {
     }
 };
 
-om.BinaryReader = class {
-
-    constructor(buffer) {
-        this._buffer = buffer;
-        this._length = buffer.length;
-        this._position = 0;
-        this._view = new DataView(buffer.buffer, buffer.byteOffset, buffer.byteLength);
-    }
-
-    seek(position) {
-        this._position = position >= 0 ? position : this._length + position;
-    }
-
-    skip(offset) {
-        this._position += offset;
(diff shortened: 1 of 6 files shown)


lutzroeder (changes requested):
Please add sample test model files showing the use of these changes.

xy-xlnx:
https://www.xilinx.com/bin/public/openDownload?filename=pt_face-quality_80_60_61.68M_1.4.zip
Download the zip file from this link, please test the PointsQuality_int.xmodel, which is an uncompiled xmodel.
https://www.xilinx.com/bin/public/openDownload?filename=face-quality_pt-vck190-r1.4.0.tar.gz
Please test the face-quality_pt.xmodel, which is a compiled xmodel.

lutzroeder:
The xmodel specific UI changes are not applied or unified to all other supported formats. Maybe this should be broken up into separate changes starting with something simple like data types. For example, the argument and data type change is not following the `int32`, `uint32` notation used for all other formats, sidestepping `TensorType` which is used for all other formats. It is unclear or not documented how `xint` differs from regular `int`.

PR #834, opened 2021-10-28
TensorFlow: Visualize missing functions in pbtxts

szymak:
Hi Lutz,

Recently, I've tried reducing the size of dumped graphs by TensorFlow, by not storing functions data in the files.
Instead, each function, since it's a self-contained graph, is stored as a separate file. 
This reduces the size of files containing nodes with functions significantly, as functions can describe very complex subgraphs.

Netron is not visualizing the content of functions currently, or at least I could not find it in the GUI.
In any case, I propose fixing it by printing the function names as string type + 'not found' suffix, so that such graphs can still be visualized.

Let me know, what you think.

Cheers,
Michał

#838

### DIFF
--- a/source/tf.js
+++ b/source/tf.js
@@ -1041,7 +1041,9 @@ tf.Attribute = class {
                 this._type = 'function';
                 this._value = metadata.type(name);
                 if (!this._value) {
-                    throw new tf.Error("Unknown function '" + name + "'.");
+                    // Function was not found in the pbtxt, but we can still visualize the graph
+                    this._type = 'string';
+                    this._value = name + " (not found)";
                 }
                 break;
             }


lutzroeder (changes requested):
@szymak can you provide the function file as well and how it is loaded for runtime execution. For example, is there a `cluster_function_1.pbtxt` with a predictable name or is this only known in client code?

szymak:
There are no 'function files'. Functions are generated from **tensorflow::Graphs**.
By storing them in separate files, I meant, I store the graphs they originated from (with the same API call [tensorflow::DumpGraphToFile](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/dump_graph.h#L51)).

TF provides way of transforming graphs into FunctionDef (using [tensorflow::GraphToFunctionDef(...)](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/graph_to_functiondef.h#L51)).
FunctionDefs are added  to FunctionLibraryDefinition via [AddFunctionDef(...)](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/function.h#L428) and later referred by names.

At the end of graph_with_function.pbtxt (inside zip), there is **library {  function { ... } }**.
It appear there, when you pass non-nullptr FunctionLibraryDefinition to [tensorflow::DumpGraphToFile(...)](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/dump_graph.h#L51).

Transformation of FunctionDef back to tensorflow::Graph is done via [InstantiateFunction(...)](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/function.h#L260) && [ConvertNodeDefsToGraph(...)](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/graph_constructor.h#L64) API calls.

In short, I do not need to have **library {  function { ... } }** present in pbtxt to be able to process the graph.
I can mark given function 'missing/not found' and still visualize the main graph.

PR #841, opened 2021-11-13
Circle: Add *.circle extension handling

glistening:
Thank you for adding `circle` support in recent version.

I've found it doesn't show `*.circle` in open file dialog.
I have to write "*.circle" to see `*.circle`.
Could you please add this change?

It also has other changes about `*.circle` support.

- Update electron-builder.yml for *.circle association
- Update description in `setup.py`.


lutzroeder:
This was left out intentionally. It is not clear that maintaining and promoting another duplicate format is the right approach. Currently there seem to be no circle files in circulation so wondering if the time is better spent on `tflite` support which will lead to less code duplication?

Issue #835, opened 2021-10-29
Netron does not display dynamic shapes along the lines


fefe982:
- Netron app and version: 
5.3.1 destop installer version
- OS and browser version: 
Windows 10

Steps to Reproduce:
Netron only displays shape with fixed demension (like `2 x 768`), but does not display dynamic dimensions （like `batch x sequence x 768`)

Please attach or link model files to reproduce the issue if necessary.

fefe982:
I created a simple onnx graph, the this is the graph rendered by v5.2.7:
<img width="387" alt="pic_5 2 7" src="https://user-images.githubusercontent.com/3244578/139617710-c1aceaee-dd04-40b6-b04e-4a8da2c8fbe6.png">
This is the one rendered by v5.3.3. All dynamic dimensions are missing here 
<img width="387" alt="pic_5 3 3" src="https://user-images.githubusercontent.com/3244578/139617716-6de74985-a73c-4c27-9f03-2c6916f3ae9d.png">
This zip file contains the onnx file and both pictures.
[netron_bug.zip](https://github.com/lutzroeder/netron/files/7450607/netron_bug.zip)

fefe982:
Thank you for fixing this.

However, I see there is a length limit of 16. The actual content can easily exceed 16 when we use meaning names for the dimension, for example: "batch x sequence x 768", or "2 x batch x 12 x sequence x 64". There are many similar shapes in my actual graph.

This limitation does not exist prior to 5.2.7

PR #846, opened 2021-11-25
remove length limit on displayed shape

fefe982:
This continues #835.

Netron now displays the shapes, but have length limit of 16 for the whole string. This limit did not exist before v5.2.7.

In my real graphs, when I want to use some meaning for names for the dimensions, like "batch", "sequence", the length of for the entire shape can easily exceed that limit, like "batch x 12 x sequence x 64". It get even long in complex graphs.

So here I propose to remove that limit.


lutzroeder:
The tradeoff is longer strings impact the layout of complex graphs. This will require some more investigation.

Issue #759, opened 2021-06-27
nnabla support


lutzroeder:
[sony/nnabla](https://github.com/sony/nnabla)

[tecogan_model.h5](https://nnabla.org/pretrained-models/nnabla-examples/GANs/tecogan/tecogan_model.h5)
[pretrained_fomm_params.h5](https://nnabla.org/pretrained-models/nnabla-examples/GANs/first-order-model/pretrained_fomm_params.h5)

PR #880, opened 2022-02-16
Add Neural Network Libraries prototype

JohnBowidowicz:
Hello lutzroeder,

I have extended Netron to support Neural Network Libraries ([nnabla](https://github.com/sony/nnabla)), which is a machine learning framework developed by Sony. Example models can be found [here](https://nnabla.readthedocs.io/en/latest/python/api/models.html).

#759

Thank you for making such an amazing tool.


lutzroeder (changes requested):
1. `./tools/nnabla sync schema` should generate the `nnabla-proto.js` file. This will probably require adding the most recent expanded `.proto` file to the nnabla repo as doing the template expansion on the fly would add too much complexity or Python dependencies.
2. `./tools/nnabla sync metadata` should generate the `nnabla-metadata.json` file.
3. `./test/models.json` should contain test files with `node ./test/models.js nnabla` passing.

tools/nnabla
```diff
@@ -0,0 +1,77 @@
...
+    venv
+    ${python} -m pip install --quiet --upgrade six mako pyyaml
+    deactivate
+}
+
+schema() {
+    echo "nnabla schema"
+    venv
+    export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
+    ${python} ./tools/nnabla-script.py proto
```
lutzroeder, line 47: There shouldn't be a requirement on Python or `mako` to generate the schema file. Netron supports dozens of formats and shouldn't be blocked on complex build steps. Ideally the latest expanded version of the `.proto` file should be provided by the nnabla repo.

lutzroeder (changes requested):

Issue #895, opened 2022-03-31
TensorFlow: object attribute support


zvika-adler:
- OS and browser version: Ubuntu20 + Chrome 97.0.4692.99
- Netron version 5.6.6

Steps to Reproduce:

1. Take attached file and change extension to .pbtxt (this was changed due to GitHub restrictions). This model is build by Protobuf GraphDef structure that enables func structure. 
2. Load with Web Netron
3. Press on Placeholder node to view attributes.

Current behavior: OutTensorInfo parameter is of type func and cannot be viewed. If pressing OutTensorInfo link, it doesn't show the data.
Expected behavior: OutTensorInfo parameter can be shown and collapses/de-collapses after clicking "+" button.

Please attach or link model files to reproduce the issue if necessary.
[myModel.txt](https://github.com/lutzroeder/netron/files/8389336/myModel.txt)

lutzroeder:
[netron_issue_895.pbtxt.zip](https://github.com/lutzroeder/netron/files/8459475/netron_issue_895.pbtxt.zip)

DoDoENT:
In general, it would suffice to simply display TF func's as key-value maps, as [this is the way they are written in the protobuf file](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/attr_value.proto#L46).

Furthermore, we also use [list of funcs](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/attr_value.proto#L40) for some of our models, so it would be great if this was supported too.

PR #896, opened 2022-04-05
support for tf func

zvika-adler:
I made the relevant chnages for tf. I didn't find any other formats that use func as a type, so I changed it it the viewer as well. If i am wrong, let me know please.


zvika-adler:
It looks like you overrode all my changes and they are not there anymore.
Is there any problem with my solution?
It should fix https://github.com/lutzroeder/netron/issues/895
Thanks

lutzroeder:
@zvika-adler trying to find an approach that works for all formats. The general case of nested object graphs likely will require a different UI.

zvika-adler:
Do I need to do something to make it progress?

zvika-adler:
I see that it is still not merged and fixed with the other formats. Is it something I can help or it may take longer than expected? @lutzroeder

DoDoENT:
I've built the app from your branch and I can confirm it works correctly for me.

DoDoENT:
I've added a [simple hack to allow visualizing tensors within functions](https://github.com/microblink/netron/commit/adcb74378940dcac1b16291e2bbcc6451fdbc610). Not a pretty, but works for my case (it's still better than not seeing anything).

Issue #861, opened 2022-01-01
Export as PNG or SVG doesn't work


zjeffer:
- Netron app and version: Netron-5.4.9, the AppImage
- OS and browser version: `Arch Linux 5.15.12`, with `bspwm 0.9.10-33-ge22d0fa`

### Steps to Reproduce:

1. Load a model (I tried it with my own models and the SqueezeNet 1.0-3 sample model from README.md)
2. Go to File > Export
3. A file manager popup opens
4. Give a name to the file
5. Choose PNG or SVG
6. Click Export

The file manager popup closes and nothing happens. No error, and no output when opening the AppImage from the commandline. There is no file saved in the specified folder.

### Expected results

A PNG or SVG file should appear in the chosen folder.

### Please attach or link model files to reproduce the issue if necessary.

Model: https://media.githubusercontent.com/media/onnx/models/master/vision/classification/squeezenet/model/squeezenet1.0-3.onnx

lutzroeder:
@zjeffer does adding a file extension to the file name resolve this?

lutzroeder:
Duplicate of electron/electron#21935

zjeffer:
> @zjeffer does adding a file extension to the file name resolve this?

@lutzroeder That works. I guess the issue is upstream, but they have no intention of fixing this: https://github.com/electron/electron/pull/15510

Is there no way to at least show the user an error when no extension is given, maybe in an `else` statement after this `if`: https://github.com/lutzroeder/netron/blob/main/source/view.js#L667

Because currently, there is no way to tell that it went wrong...

lutzroeder:
@zjeffer @shawnmasters please submit a pull request with improvements. The fix would have to be scoped to `_export()` in `app.js` to be scoped to Linux and the desktop app.

PR #986, opened 2022-10-05
Show error dialog when filename does not have an extension.

zjeffer:
This is a quick and dirty fix for #861. I haven't found how to automatically add the extension to the filename when it doesn't have one, because I can't find how to get the extension from the type selector in the showSaveDialog box. 

Either way, this should really be fixed upstream instead, but this is just a minimal fix so the user can at least know that the file didn't save.


zjeffer:
Also this should be tested on other platforms obviously.

PR #991, opened 2022-10-14
Load scripts in parallel to reduce total page load time by ~2 seconds

josephrocca:
I noticed that the loading was a bit slow due to the serial loading of the scripts:

https://user-images.githubusercontent.com/1167575/195853315-650fb495-4e5b-4d69-8049-2b6555cf2007.mp4

This patch should cut load time down from ~4 seconds to ~2 seconds. It assumes that all scripts other than `index.js` can be loaded in parallel.

Please feel free to close this pull request if you'd rather implement this in a different way (e.g. by bundling the scripts).

Thanks for all your work on this project!

### DIFF
--- a/source/index.html
+++ b/source/index.html
@@ -21,7 +21,7 @@
                 let scriptEl = document.createElement("script");
                 scriptEl.onload = resolve;
                 scriptEl.src = url;
-                document.body.appendChild(scriptEl);
+                document.head.appendChild(scriptEl);
             });
         }
         let urls = ["dagre.js","base.js","text.js","json.js","xml.js","python.js","protobuf.js","flatbuffers.js","flexbuffers.js","zip.js","gzip.js","tar.js","view-grapher.js","view-sidebar.js","view.js"];
--- a/source/index.html
+++ b/source/index.html
@@ -14,22 +14,23 @@
 <link rel="apple-touch-icon" type="image/png" href="icon.png">
 <link rel="apple-touch-icon-precomposed" type="image/png" href="icon.png">
 <link rel="fluid-icon" type="image/png" href="icon.png">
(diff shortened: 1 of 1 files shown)


lutzroeder:
image

josephrocca:
Ahh, it should have been `document.head.appendChild` rather than `document.body.appendChild`. I've made that change, but I now realise that there are dependencies between the scripts. E.g. at the top of `gzip.js` there's this line:
```js
var zip = zip || require('./zip');
```
And since we're now loading the scripts in parallel it's possible for `gzip.js` to load before `zip.js` which will cause an error on the web saying "`require` is not defined". Does `zip` need to be assigned synchronously here? If not you could move all those assignments to the top of `index.js`. If you use `var` during assignment, then that will make them available globally - i.e. to all scripts, no matter which script you declared them in.

If those assignments are needed synchronously (i.e. the immediate execution of the `gzip.js` script as the page loads requires the `zip` variable to be assigned), then either a refactor may be needed, or you could load the scripts in a sequence of several batches, based on their dependencies. If you have a long sequence of dependencies then that might defeat the purpose of this change.

Ideally you don't need the assignments like `var zip = zip || require('./zip');` until you get to `index.js`, so you could just move them all to the top of `index.js`.

lutzroeder:
> Ideally you don't need the assignments like var zip = zip || require('./zip'); until you get to index.js, so you could just move them all to the top of index.js.

Required for the Electron desktop app.

josephrocca:
@lutzroeder Could you do something like this?

```
if (typeof window === 'undefined') { // i.e. only in Electron
  var zip = zip || require('./zip');
}
```

The `var` makes `zip` hoisted, so it will be available as a global within the script in Electron.

Issue #993, opened 2022-10-18
DaVinci: Unsupported file content for extension '.om' 


wujunliang00:
Hello @lutzroeder,
I encounterd this issue when trying to open the om file. The file format is svp type for HiAi，and netron does not support opening.
<img width="432" alt="3de1d1ae4278b25950e961fe58d5261" src="https://user-images.githubusercontent.com/42567887/196493959-9f2d0bc3-f743-4dfe-afc3-e804185163af.png">

[resnet50_issue_svp.zip](https://github.com/lutzroeder/netron/files/9813495/resnet50_issue_svp.zip)

PR #994, opened 2022-10-18
DaVinci : fix svp format support

wujunliang00:
Hi, I'm trying to support DaVinci model(.om) svp format in netron, can u help me review the code and merge the new feature into the main branch? 
thanks!

the sample models:
[resnet50_issue_svp.zip](https://github.com/lutzroeder/netron/files/9813559/resnet50_issue_svp.zip)
Also,  this request can be solved https://github.com/lutzroeder/netron/issues/993

### DIFF
--- a/source/python.js
+++ b/source/python.js
@@ -1619,11 +1619,11 @@ python.Tokenizer = class {
 
 python.Execution = class {
 
-    constructor(sources, exceptionCallback) {
+    constructor(sources) {
         const self = this;
         const execution = self;
         this._sources = sources || new Map();
-        this._exceptionCallback = exceptionCallback;
+        this._events = new Map();
         this._utf8Decoder = new TextDecoder('utf-8');
         this._unresolved = new Map();
         const dict = class extends Map {};
@@ -5455,7 +5455,7 @@ python.Execution = class {
             if (!this._unresolved.has(name)) {
                 const moduleName = name.split('.').shift();
                 if (this._registry.has(moduleName) && moduleName !== '__main__') {
-                    this._exceptionCallback(new python.Error("Unknown type name '" + name + "'."), false);
+                    this.emit('resolve', name);
                 }
                 const type = this._createType(name, class {});
                 this._unresolved.set(name, type);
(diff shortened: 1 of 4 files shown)


lutzroeder:
@wujunliang00 
* Why is there a different format, should to `.proto` files for both formats get merged and unified?
* The code in `om-proto-svp.js` looks handed edited. Add the `.proto` in `./tools/om` and regenerate the file.
* All `.proto` definitions should get merged into `om-proto.js` .

lutzroeder (changes requested):
@wujunliang00

* Why is there a different format, should the `.proto` files for both formats get merged and unified?
* The code in `om-proto-svp.js` looks handed-edited. Add the `.proto` in `./tools/om` and regenerate the file.
* The `.proto` definitions for `om.js` should be in `om-proto.js`.
* Who will maintain and support this change going forward?

wujunliang00:
> @wujunliang00
> 
> * Why is there a different format, should the `.proto` files for both formats get merged and unified?
> * The code in `om-proto-svp.js` looks handed-edited. Add the `.proto` in `./tools/om` and regenerate the file.
> * The `.proto` definitions for `om.js` should be in `om-proto.js`.
> * Who will maintain and support this change going forward?

* Since this structure is a model on the deployment side of HiAi, in order to make the performance faster during deployment, it adopts its own parsing, which makes it impossible to parse in the form of protobuf.
* Due to the use of another parsing method, it is not compatible with `om-proto`, sorry~~ , I don't have the ability to be compatible on `om.proto`.
*  HiAscend maintains and support this change going forward, please refer to this: https://www.hiascend.com/en/
* The OM model of 'PICO' is generally deployed on platforms with low computing power.

lutzroeder:
@wujunliang00

* Calling the `om-proto-svp` is misleading if it not using Protocol Buffers
* What does the `svp` moniker stand for?
* Is there an overview of the different DaVinci OM formats?
* Please add support for weight initializers

wujunliang00:
> @wujunliang00
> 
> * What does the `svp` moniker stand for?
> * Is there an overview of the different DaVinci OM formats?
> * Please add support for weight initializers

1.  svp：smart vision pico of om
2. At present, I am adapting to the om of svp, and there are other types for DaVinci OM, but they are all in the direct proto format, and do not need original parsing like me
3. weight initializers will be adapted in the future, and it is not supported for the time being.

lutzroeder:
@wujunliang00 is there a reason to not add weight initializers now? It is the main reason to keep the feature disabled.

wujunliang00:
@lutzroeder 

Because the weight data is placed in the instructions issued by the hardware, it is convenient to directly deploy the om file on the hardware, but it is very difficult to parse natively, because the weight data has been converted into instructions and stored in om

PR #995, opened 2022-10-18
Add initial implementation for protobuf binary encoding

weimingzha0:


weimingzha0:
This is my first PR and my first JS code, so before I implement the full functionality, it's better to make sure it is on the right track. :)

weimingzha0:
Hi @lutzroeder will this PR be merged?

lutzroeder:
Ideally there should be a feature using the code for review, to avoid merging code that later isn't tested or used. Model editing is a tricky one, supporting it for one format will likely set expectations that you are working to support and maintain it for many other formats.

weimingzha0:
yes, I agree that it might set some expectations to support model editing. 
To lower the expectation, we can start from low-level libs like pb encoding or other encoding.
When low-level libs are ready for writing, we can further design how to enable the editing via web UI.

PR #1005, opened 2022-11-21
oneDNN support

xinyu-intel:
Hi, I'm trying to support [oneDNN](https://github.com/oneapi-src/oneDNN) serialized graph(JSON format) in netron, can you help me review the code and merge the new feature into the main branch?
thanks!

the sample graph:
[sample-graph-100001.json.zip](https://github.com/lutzroeder/netron/files/10037438/sample-graph-100001.json.zip)

resnet50:
[resnet50-sample-graph.json.zip](https://github.com/lutzroeder/netron/files/10065168/resnet50-sample-graph.json.zip)

About oneDNN Graph API: https://www.oneapi.io/blog/oneapi-extended-with-graph-api-to-simplify-integration-and-maximize-computation-efficiency/

Spec: https://spec.oneapi.io/versions/1.2-rev-1/elements/oneDNN/source/graph/index.html

#1006

### DIFF
--- a/source/grapher.css
+++ b/source/grapher.css
@@ -37,6 +37,7 @@
 .node-item-type-data path { fill: rgb(85, 85, 85); }
 .node-item-type-quantization path { fill: rgb(80, 40, 0); }
 .node-item-type-custom path { fill: rgb(128, 128, 128); }
+.node-item-type-onednngraph-backend0 path { fill: rgb(0, 113, 197); }
 
 .node-item-input path { fill: #fff; }
 .node-item-input:hover { cursor: pointer; }
--- a/source/grapher.css
+++ b/source/grapher.css
@@ -37,7 +37,6 @@
 .node-item-type-data path { fill: rgb(85, 85, 85); }
 .node-item-type-quantization path { fill: rgb(80, 40, 0); }
 .node-item-type-custom path { fill: rgb(128, 128, 128); }
-.node-item-type-onednngraph-backend0 path { fill: rgb(0, 113, 197); }
 
 .node-item-input path { fill: #fff; }
 .node-item-input:hover { cursor: pointer; }
--- a/source/onednngraph.js
+++ b/source/onednngraph.js
@@ -1,0 +1,515 @@
+
+var onednngraph = onednngraph || {};
+var json = json || require('./json');
+var base = base || require('./base');
+
+onednngraph.ModelFactory = class {
+
+    match(context) {
+        const identifier = context.identifier;
+        const extension = identifier.split('.').pop().toLowerCase();
+        if (extension === 'json') {
+            const obj = context.open('json');
+            if (obj && obj.version && obj.graph) {
+                return 'onednngraph.json';
+            }
+        }
+        return undefined;
+    }
+
+    open(context, match) {
+        return onednngraph.Metadata.open(context).then((metadata) => {
+            const createModel = (metadata, symbol, version) => {
+                return new onednngraph.Model(metadata, symbol, version);
+            };
+            switch (match) {
+                case 'onednngraph.json': {
+                    let symbol = null;
+                    try {
+                        symbol = context.open('json');
+                    }
+                    catch (error) {
+                        const message = error && error.message ? error.message : error.toString();
+                        throw new onednngraph.Error("Failed to load symbol entry (" + message.replace(/\.$/, '') + ').');
+                    }
+                    const version = symbol && symbol.version ? symbol.version : null;
+                    return createModel(metadata, symbol, version);
+                }
+                default: {
+                    throw new onednngraph.Error("Unsupported oneDNN Graph format '" + match + "'.");
+                }
+            }
+        });
+    }
+};
+
+onednngraph.Model = class {
+
+    constructor(metadata, symbol, version) {
+        if (!symbol) {
+            throw new onednngraph.Error('JSON symbol data not available.');
+        }
+        if (symbol) {
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'graph')) {
+                throw new onednngraph.Error('JSON file does not contain an oneDNN Graph \'graph\' property.');
+            }
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'version')) {
+                throw new onednngraph.Error('JSON file does not contain an oneDNN Graph \'version\' property.');
+            }
+        }
+        this._format = 'oneDNN Graph';
+        this._version = 'oneDNN' + (version ? ' v' + version : '');
+        this._license = 'Apache License 2.0';
+        this._graphs = [new onednngraph.Graph(metadata, symbol)];
+    }
+
+    get format() {
+        return this._format;
+    }
+
+    get producer() {
+        return this._producer;
+    }
+
+    get name() {
+        return this._name;
+    }
+
+    get version() {
+        return this._version;
+    }
+
+    get description() {
+        return this._description;
+    }
+
+    get author() {
+        return this._author;
+    }
+
+    get license() {
+        return this._license;
+    }
+
+    get runtime() {
+        return this._runtime;
+    }
+
+    get graphs() {
+        return this._graphs;
+    }
+};
+
+onednngraph.Graph = class {
+
+    constructor(metadata, symbol) {
+        this._metadata = metadata;
+        this._nodes = [];
+        this._inputs = [];
+        this._outputs = [];
+
+        if (symbol) {
+            this._type = 'engine: ' + symbol.engine_kind + '; fpmath: ' + symbol.fpmath_mode + ';';
+            const nodes = symbol.graph;
+
+            const args = new Map();
+            const arg = (name, type, lt) => {
+                if (!args.has(name)) {
+                    args.set(name, new onednngraph.Argument(name, type, null, lt));
+                }
+                return args.get(name);
+            };
+            for (const node of nodes) {
+                this._nodes.push(new onednngraph.Node(this._metadata, node, arg, symbol.engine_kind));
+            }
+        }
+    }
+
+    get name() {
+        return '';
+    }
+
+    get type() {
+        return this._type;
+    }
+
+    get inputs() {
+        return this._inputs;
+    }
+
+    get outputs() {
+        return this._outputs;
+    }
+
+    get nodes() {
+        return this._nodes;
+    }
+};
+
+onednngraph.Node = class {
+
+    constructor(metadata, node, arg, device) {
+        const type = node.kind;
+        this._name = node.name;
+        this._attributes = [];
+        this._inputs = [];
+        this._outputs = [];
+        this._type = { name: type };
+        this._device = device;
+        this._description = 'oneDNN Graph OP with unique id: ' + node.id;
+
+        const attrs = node.attrs;
+        if (attrs) {
+            for (const attributeName of Object.keys(attrs)) {
+                this._attributes.push(new onednngraph.Attribute(metadata, type, attributeName, attrs[attributeName].type, attrs[attributeName].value));
+                if (attributeName == 'backend') {
+                    if (attrs[attributeName].value == 'dnnl_backend') {
+                        this._type.category = "onednngraph-backend0";
+                    }
+                    else {
+                        // leave color of fake_backend node by default
+                    }
+                }
+            }
+        }
+
+        const inputs = node.inputs || [];
+        let inputIndex = 0;
+        for (const input of inputs) {
+            if (inputIndex < inputs.length) {
+                const inputArguments = [];
+                const lt = new onednngraph.LogicalTensor(input.id, input.dtype, input.shape, input.stride, input.layout_type, input.property_type);
+                const shape = new onednngraph.TensorShape(lt.shape);
+                const type = new onednngraph.TensorType(lt.dataType, shape);
+                inputArguments.push(arg(input.id.toString(), type, lt));
+                const inputName = (inputs.length == 1) ? 'input' : ('input' + (inputIndex)).toString();
+                this._inputs.push(new onednngraph.Parameter(inputName, inputArguments));
+                inputIndex += 1;
+            }
+        }
+        this._inputs.push(...inputs.slice(inputIndex).map((input, index) => {
+            const inputName = ((inputIndex + index) == 0) ? 'input' : (inputIndex + index).toString();
+            return new onednngraph.Parameter(inputName, true, [arg(input)]);
+        }));
+
+        const outputs = node.outputs || [];
+        let outputIndex = 0;
+        for (const output of outputs) {
+            if (outputIndex < outputs.length) {
+                const outputArguments = [];
+                const lt = new onednngraph.LogicalTensor(output.id, output.dtype, output.shape, output.stride, output.layout_type, output.property_type);
+                const shape = new onednngraph.TensorShape(lt.shape);
+                const type = new onednngraph.TensorType(lt.dataType, shape);
+                outputArguments.push(arg(output.id.toString(), type, lt));
+                const outputName = (outputs.length == 1) ? 'output' : ('output' + (outputIndex)).toString();
+                this._outputs.push(new onednngraph.Parameter(outputName, outputArguments));
+                outputIndex += 1;
+            }
+        }
+        this._outputs.push(...outputs.slice(outputIndex).map((output, index) => {
+            const outputName = ((outputIndex + index) == 0) ? 'output' : (outputIndex + index).toString();
+            return new onednngraph.Parameter(outputName, true, [arg(output)]);
+        }));
+    }
+
+    get type() {
+        return this._type;
+    }
+
+    get name() {
+        return this._name;
+    }
+
+    get inputs() {
+        return this._inputs;
+    }
+
+    get outputs() {
+        return this._outputs;
+    }
+
+    get attributes() {
+        return this._attributes;
+    }
+
+    get description() {
+        return this._description;
+    }
+
+    get device() {
+        return this._device;
+    }
+};
+
+onednngraph.Attribute = class {
+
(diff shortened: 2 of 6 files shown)


source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
+
+var onednngraph = onednngraph || {};
```
lutzroeder, line 2: `var onednn = onednn || {};`

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+                    let symbol = null;
+                    try {
+                        symbol = context.open('json');
+                    }
+                    catch (error) {
+                        const message = error && error.message ? error.message : error.toString();
+                        throw new onednngraph.Error("Failed to load symbol entry (" + message.replace(/\.$/, '') + ').');
+                    }
+                    const version = symbol && symbol.version ? symbol.version : null;
+                    return createModel(metadata, symbol, version);
```
lutzroeder, line 36: Inline as it is called only once.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+            };
+            switch (match) {
+                case 'onednngraph.json': {
+                    let symbol = null;
+                    try {
+                        symbol = context.open('json');
+                    }
+                    catch (error) {
+                        const message = error && error.message ? error.message : error.toString();
+                        throw new onednngraph.Error("Failed to load symbol entry (" + message.replace(/\.$/, '') + ').');
```
lutzroeder, line 33: `context.open` should always succeed and `catch` is not needed?

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+};
+
+onednngraph.Model = class {
+
+    constructor(metadata, symbol, version) {
+        if (!symbol) {
+            throw new onednngraph.Error('JSON symbol data not available.');
+        }
+        if (symbol) {
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'graph')) {
```
lutzroeder, line 53: `if (symbol.graph) {`

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+var base = base || require('./base');
+
+onednngraph.ModelFactory = class {
+
+    match(context) {
+        const identifier = context.identifier;
+        const extension = identifier.split('.').pop().toLowerCase();
+        if (extension === 'json') {
+            const obj = context.open('json');
+            if (obj && obj.version && obj.graph) {
```
lutzroeder, line 13: Multiple JSON formats can define `version` and `graph` properties, add unique approach to identify the format?

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'graph')) {
+                throw new onednngraph.Error('JSON file does not contain an oneDNN Graph \'graph\' property.');
+            }
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'version')) {
+                throw new onednngraph.Error('JSON file does not contain an oneDNN Graph \'version\' property.');
+            }
+        }
+        this._format = 'oneDNN Graph';
+        this._version = 'oneDNN' + (version ? ' v' + version : '');
+        this._license = 'Apache License 2.0';
```
lutzroeder, line 62: Remove. This should return the license of the model as defined by the model author.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+        }
+        if (symbol) {
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'graph')) {
+                throw new onednngraph.Error('JSON file does not contain an oneDNN Graph \'graph\' property.');
+            }
+            if (!Object.prototype.hasOwnProperty.call(symbol, 'version')) {
+                throw new onednngraph.Error('JSON file does not contain an oneDNN Graph \'version\' property.');
+            }
+        }
+        this._format = 'oneDNN Graph';
```
lutzroeder, line 60: `this._format = 'oneDNN Graph' + (version ? ' v' + version : '');`

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+        this._license = 'Apache License 2.0';
+        this._graphs = [new onednngraph.Graph(metadata, symbol)];
+    }
+
+    get format() {
+        return this._format;
+    }
+
+    get producer() {
+        return this._producer;
```
lutzroeder, line 71: Remove properties which are not used.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+onednngraph.Graph = class {
+
+    constructor(metadata, symbol) {
+        this._metadata = metadata;
+        this._nodes = [];
+        this._inputs = [];
+        this._outputs = [];
+
+        if (symbol) {
+            this._type = 'engine: ' + symbol.engine_kind + '; fpmath: ' + symbol.fpmath_mode + ';';
```
lutzroeder, line 112: Use `model.runtime` property.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+
+    constructor(metadata, node, arg, device) {
+        const type = node.kind;
+        this._name = node.name;
+        this._attributes = [];
+        this._inputs = [];
+        this._outputs = [];
+        this._type = { name: type };
+        this._device = device;
+        this._description = 'oneDNN Graph OP with unique id: ' + node.id;
```
lutzroeder, line 159: Delete. What is `node.id` and why do noes have an `id` and `name`? If it is an index, maybe use `location`.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+
+        const attrs = node.attrs;
+        if (attrs) {
+            for (const attributeName of Object.keys(attrs)) {
```
lutzroeder, line 163: `Object.entries` instead of `Object.keys` to avoid duplicate lookups?

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+    get device() {
+        return this._device;
+    }
+};
+
+onednngraph.Attribute = class {
+
+    constructor(metadata, type, name, attr_type, attr_value) {
+        this._name = name;
+        this._type = attr_type;
```
lutzroeder, line 248: `type` should return the same types as all other formats. For example, `boolean`, `int32`, `uint32`, `uint32[]`.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+        }
+        return this._type;
+    }
+
+    get initializer() {
+        return this._initializer;
+    }
+
+    get description() {
+        const stride = this._lt.stride ? ('[' + this._lt.stride.map((stride) => stride ? stride.toString() : '?').join(',') + ']') : '';
```
lutzroeder, line 383: Use `TensorType.denotation` or `Tensor.stride`. `description` should be an actual description provided by the model author, not a generated string.

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+    get dimensions() {
+        return this._dimensions;
+    }
+
+    toString() {
+        return this._dimensions ? ('[' + this._dimensions.map((dimension) => dimension ? dimension.toString() : '?').join(',') + ']') : '';
+    }
+};
+
+onednngraph.LogicalTensor = class {
```
lutzroeder, line 425: `onednn.Tensor`. Why makes a tensor logical?

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+    get layout_type() {
+        return this._layout_type;
+    }
+
+    get property_type() {
+        return this._property_type;
+    }
+};
+
+onednngraph.Metadata = class {
```
lutzroeder, line 461: Delete. This code is not used. If needed, `context.metadata(...)` will provide this standard implementation.

lutzroeder (changes requested):
- [x] Add real models file to `./test/models.json` and run `node ./test/models.js onednn`
- [x] Add metadata for category and inputs names for common types like `Convolution` or `ReLU`.

source/grapher.css
```diff
@@ -37,6 +37,7 @@
 .node-item-type-data path { fill: rgb(85, 85, 85); }
 .node-item-type-quantization path { fill: rgb(80, 40, 0); }
 .node-item-type-custom path { fill: rgb(128, 128, 128); }
+.node-item-type-onednngraph-backend0 path { fill: rgb(0, 113, 197); }
```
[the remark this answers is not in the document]
xinyu-intel: removed. thx for reminder.

source/view.js
```diff
@@ -1623,6 +1623,7 @@ view.ModelFactoryService = class {
         this.register('./hickle', [ '.h5', '.hkl' ]);
         this.register('./nnef', [ '.nnef', '.dat' ]);
         this.register('./cambricon', [ '.cambricon' ]);
+        this.register('./onednngraph', [ '.json'] );
```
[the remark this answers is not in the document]
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: changed

source/onednngraph.js
```diff
@@ -0,0 +1,515 @@
...
+
+onednngraph.ModelFactory = class {
+
+    match(context) {
+        const identifier = context.identifier;
+        const extension = identifier.split('.').pop().toLowerCase();
+        if (extension === 'json') {
+            const obj = context.open('json');
+            if (obj && obj.version && obj.graph) {
+                return 'onednngraph.json';
```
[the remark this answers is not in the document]
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: yes, changed.

source/onednngraph.js
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: added engine_kind and fpmath_mode as identifier.

source/onednngraph.js
xinyu-intel: got it.

source/onednngraph.js
xinyu-intel: changed.

source/onednngraph.js
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: changed. id act as an unique index. name can be non-unique.

source/onednngraph.js
xinyu-intel: good catch.

source/onednngraph.js
xinyu-intel: changed

source/onednngraph.js
xinyu-intel: changed to stride. Any ideas for `layout_type(strided/any/undef)` and `property_type(constant/variable/undef)`?

source/onednngraph.js
xinyu-intel: removed. Actually, it is a simple wraper for meta info of onednn graph tensor. These info can be directly feed into TensorShape and TensorType.

source/onednngraph.js
xinyu-intel: changed.

source/onednngraph.js
xinyu-intel: should be `this._format = 'oneDNN Graph' + (version ? ' v' + version[0] : '');`? Becase the format of v3.x.x will be backward compatible.

source/onednngraph.js
lutzroeder: Looks like it would be more straightforward to write the `onndnn.Attribute` constructor to translate instead of the the `switch (schema.type)` code?

source/onednngraph.js
lutzroeder: `this._runtime = symbol.engine_kind + ' ' + symbol.fpmath_mode`?

source/onednngraph.js
lutzroeder: Remove `this._version = 'oneDNN' + (version ? ' v' + version : '');` as well, is part of `this._format`.

source/onednngraph.js
lutzroeder: `onednn.Tensor` will be needed to show the initial value of weights. For example, `Convolution.inputs[1]` should have a weights value that is exposed as `onednn.Argument.initializer`.

source/onednngraph.js
lutzroeder: `constant` is `onednn.Argument` exposing `initializer`. The `initializer` is a `onednn.Tensor` object which exposes a `stride` property if the tensor is strided.

source/onednngraph.js
xinyu-intel: good idea.

source/onednngraph.js
lutzroeder: `if (!Object.prototype.hasOwnProperty.call(symbol, 'graph')) {` will never evaluate to `true`.

source/onednngraph.js
xinyu-intel: got your point. `property_type ` in onednn means whether a tensor is constant. For example, given weight of  convolution(a.k.a input[1]), it can be `Variable` in a training graph and `constant` in a inference graph.

source/onednngraph.js
lutzroeder: For example, "Wildcard"-style nodes in other formats get rendered as part of the `argument.initializer` in other formats. This requires the `initializer` to be set even when empty. It might also help to differentiate between `constant` and `variable` nodes.

source/onednngraph.js
lutzroeder: Is the `switch (schema.type) {` code in `onednn.Attribute` actually required. The format seems to include the attribute name and type in the file so the code transforming attribute values based on types could be merged into `switch (attr_type) {`.

source/onednngraph.js
lutzroeder: `this._runtime = obj.engine_kind + ' ' + obj.fpmath_mode;` should be used in `onednn.Model` constructor instead of `this._runtime = 'engine: ' + symbol.engine_kind + '; fpmath: ' + symbol.fpmath_mode + ';';`.

lutzroeder: Is the version check `if (!Object.prototype.hasOwnProperty.call(symbol, 'version')) {` actually required? The code seems to work without it.

source/onednngraph.js
xinyu-intel: Great. One question here is that does every inputs of operator need to set initializer? Maybe I need to take a deep understand this part since I missed this previously. BTW, is it possible to support the basic part(graph, node, attr, TensorShape, TensorType) first and then I'll enhance the Tensor part. What do you think?

source/onednngraph.js
lutzroeder: Same with `if (!symbol) {` which never evaluates to `true`.

source/onednngraph.js
lutzroeder: `onednn-metadata.json` contains invalid category names, check other metadata files for reference. `Convolution` is `Layer`. `AvgPool` is `Pool`. `lower_precision` and `misc` are not valid categories.

source/onednngraph.js
lutzroeder: Only inputs that have an initial value, usually the weights. ONNX [example](https://netron.app/?url=https://media.githubusercontent.com/media/onnx/models/main/vision/classification/squeezenet/model/squeezenet1.0-3.onnx) for reference.

xinyu-intel: Thanks for the example. I'll try to fullfill this part according to this.

source/onednngraph.js
xinyu-intel: updated and node type schema has been enabled. Now, some ops in the graph should be rendered with defined css style.

lutzroeder: Please also update the ordering of properties to match other `-metadata.json` files, e.g. `name` before `category`.

xinyu-intel: done.

lutzroeder: Order is `name`, `category`, `inputs`, ... in other files.

xinyu-intel: seems category is the last in onnx-metafile... Anyway, can change it. What't the impact of order?

source/onednngraph.js
xinyu-intel: Hi, thanks for your guide. After tried locally, I tend to not set the initializer and also removed the `layout_type` and `property_type` from description. Here are some reasons:
1) onednn graph is a lower level graph compared with framework. It is hard to represent a real `Model` with the graph. For example, the graph doesn't define any `data` or `weight`, and we can only know how many inputs/outputs for each node.
2) Every inputs/outputs of node have `property_type`, which means even for an activation, it can be constant. I've tried to set initializer for every arguments and it appears that edge info all missed.
3) I agree with you that `description` is not a suitable place to show the info. So, I decide to remove them.

lutzroeder: Sounds good.

1) The input parameter names for types like `Convolution` should be descriptive, (e.g. `input` and `weight` instead of `input0` and `input1`.

2) All other formats fold constant nodes that are used only once to be initializers to make the graph more readable. For example, `Wildcard` nodes with no inputs, a single output and not attributes could be folded into the parent node as an initializer with `category` = `Wildcard`.

xinyu-intel: done.

xinyu-intel:
@lutzroeder Thanks for the merging. Can you also help to update the gh-pages?

PR #1014, opened 2022-12-01
onednn: support more dtype and update readme

xinyu-intel:


xinyu-intel:
@lutzroeder how about readme?

lutzroeder:
Usually adding this later when bits are in use and stable. Generally, there are too many ML formats, most are similar or identical to existing ones. Ideally the list should become shorter and duplicate formats replaced with more broadly supported ones like ONNX.

PR #1022, opened 2022-12-14
MegEngine: fix some bugs

Ysllllll:
fix some bugs of megengine C++ model (.mge) visualization:

1. show the shape of the middle tensor;
2. fix scope matching model identifier (mgv2) due to possible leading information;

please help review, thanks~


source/megengine.js
```diff
@@ -8,20 +8,16 @@ megengine.ModelFactory = class {
...
                 if (reader.identifier === 'mgv2') {
-                    return 'megengine.mge';
-                }
+                    return 'megengine.flatbuffers';
             }
-            const tag = String.fromCharCode.apply(null, buffer);
-            for (const value of [ 'mgb0001', 'mgb0000a', 'MGBS', 'MGBC', 'mgbtest0' ]) {
-                if (tag.startsWith(value)) {
-                    return 'megengine.' + value;
-                }
```
lutzroeder, line 24: Why remove the code detecting other formats?

source/megengine.js
```diff
@@ -415,6 +411,10 @@ megengine.Graph = class {
                 const argument = new megengine.Argument(name, type, initializer, quantization);
                 args.push(argument);
             }
+            else if(opr.shape) {
```
lutzroeder, line 414: space after `if`

source/megengine.js
```diff
@@ -415,6 +411,10 @@ megengine.Graph = class {
...
+                const type = new megengine.TensorType('TypePlaceHolder', new megengine.TensorShape(opr.shape));
```
lutzroeder, line 415: What is `TypePlaceHolder`?

source/megengine.js
```diff
@@ -688,11 +693,7 @@ megengine.TensorType = class {
             this._dataTypeName = megengine.Utility.enum(megengine.schema, dtype_class, this._value).toLowerCase();
         }
         megengine.TensorType._dataTypeMap = megengine.TensorType._dataTypeMap || new Map([
-            [ 'bool', 'boolean' ],
-            [ 'byte', 'uint8' ], [ 'quantizeds4asymm', 'uint8' ], [ 'quantizeds8asymm', 'uint8' ], [ 'uintb4', 'uint8' ],
-            [ 'quantizeds1', 'int8' ], [ 'quantizeds4', 'int8' ], [ 'quantizeds8', 'int8' ], [ 'intb1', 'int8' ], [ 'intb2', 'int8' ], [ 'intb4', 'int8' ], [ 'qint8', 'int8' ],
-            [ 'quantizeds16', 'int16' ],
-            [ 'quantizeds32', 'int32' ]
+            [ 'bool', 'boolean' ], [ 'byte', 'uint8' ], [ 'quantizeds4asymm', 'uint8' ], [ 'quantizeds8asymm', 'uint8' ], [ 'uintb4', 'uint8' ], [ 'quantizeds1', 'int8' ], [ 'quantizeds4', 'int8' ], [ 'quantizeds8', 'int8' ], [ 'intb1', 'int8' ], [ 'intb2', 'int8' ], [ 'intb4', 'int8' ], [ 'qint8', 'int8' ], [ 'quantizeds16', 'int16' ], [ 'quantizeds32', 'int32' ], [ 'TypePlaceHolder', 'boolean' ]
```
lutzroeder, line 696: Each pair on new line to make diff easier in the future?

tools/megengine
```diff
@@ -17,6 +17,7 @@ sync() {
     curl --silent --location --output "./third_party/source/megengine/src/serialization/fbs/opr_param_defs.fbs" "https://github.com/MegEngine/MegEngine/blob/master/ci/compatibility/fbs/V2-backup/opr_param_defs.fbs?raw=true"
     curl --silent --location --output "./third_party/source/megengine/src/serialization/fbs/schema_v2.fbs" "https://github.com/MegEngine/MegEngine/blob/master/ci/compatibility/fbs/V2-backup/schema_v2.fbs?raw=true"
     sed -i 's/struct/table/g; s/\<DTypeEnum\>/fbs.DTypeEnum/g; s/\<DType\>/fbs.DType/g; s/\[ubyte:4\]/ubyte/g; s/(bit_flags)//g; s/table fbs.DType/table DType/g; s/table fbs.DTypeEnum/table DTypeEnum/g' ./third_party/source/megengine/src/serialization/fbs/*[^dtype].fbs
+    sed -i 's/items\:\[IndexDescMaskItem\]/items\:IndexDescMaskItem/g; s/desc\:\[AxisDesc\]/desc\:AxisDesc/g' ./third_party/source/megengine/src/serialization/fbs/mgb_cpp_opr.fbs
```
lutzroeder, line 20: Can you provide some context why these changes are needed? Should this get fixed in the `.fbs` file?

lutzroeder (changes requested):

source/megengine.js
Ysllllll: 👍👍 Sorry, it was deleted by mistake.

source/megengine.js
Ysllllll: done.

source/megengine.js
Ysllllll: Good idea, done.

source/megengine.js
Ysllllll: This is a modification to the first bug. Sometimes our node has no tensor information but only shape information (that is, if-else here). At this time, there will be no type, but we need to display the shape, so typeplaceholder is used here to indicate this situation.

lutzroeder: Would it work to use `'?'` as some of the other frameworks?

Ysllllll: yes, that's a good idea. done.

tools/megengine
Ysllllll: In the process of user visualization, we frequently find that there is a boundary crossing problem(i.e. `Offset is outside the bounds of the DataView`). After debugging, we find that the problem lies in these two params and is related to our dump process. Only the number of params in the dump, but not the actual content.
Since the user is using it and the information of these two params(i.e. IndexDescMaskDump and AxisAddRemove) is irrelevant, we have temporarily adjusted it so that the user no longer feels the problem. Due to the arrangement of work, we will conduct further investigation later.
Here is an example of minimal replication.

lutzroeder: Can you please investigate. Are the files corrupted or does the issue reproduce during runtime execution? Rewriting file schemas is problematic as these issues can't be addressed later without breaking compatibility.

source/megengine.js
lutzroeder: Why is this code change needed? Is there a repro example?

```javascript
var i = String(stream.peek(24)).indexOf("mgv2");
if (i >= 0) {
    i = ((i - 8) > 0 ? (i - 8): 0);
    stream.skip(i);
}
```

source/megengine.js
lutzroeder: Why is `?` mapped to `boolean` via `[ 'bool', 'boolean' ], [ '?', 'boolean' ]`? Also `this._dataTypeName` and `this._dataType` should be identical so `toString()` uses the same name as `type.dataType`?

source/megengine.js
Ysllllll: Because sometimes some leading information may be added before calling dump in python, although this is not a normal usage. For example:
$ hexdump -C  -n 100  model.mge 
00000000  6d 67 62 74 65 73 74 30  01 00 00 00 54 b8 16 08  |mgbtest0....T...|
00000010  1c 00 00 00 6d 67 76 32  14 00 24 00 20 00 1c 00  |....mgv2..$. 
```
            const tag = String.fromCharCode.apply(null, buffer);
            for (const value of [ 'mgb0001', 'mgb0000a', 'MGBS', 'MGBC', 'mgbtest0' ]) {
                if (tag.startsWith(value)) {
                    return 'megengine.' + value;
                }
```
I found that the above code was added by you not long ago. Some models do have this information at the beginning, but they use a non fbs serialization method.

source/megengine.js
Ysllllll: After testing, `['?', 'boolean']` here can be deleted.
The `this._dataType` here is mainly used to dynamically read the numerical data of the `megengine.Tensor` during visualization. It seems that the reading of the numerical data of the `*.Tensor` is no longer customized by the .js files of various frameworks? Therefore, `this._dataTypeName` is used to display the data type of mge.

source/megengine.js
lutzroeder: Does the added information have a standardized format that can be detected? What if the information is more than 24 characters or changes in the future? Can you share or add sample test files.

source/megengine.js
lutzroeder: Is there a way to generalize this? For example, can these types be added and supported for other formats as well by the shared rendering code?

source/megengine.js
Ysllllll: There seems to be no standard format for the information added. The added information is actually written in before calling the serialized interface (`dump`), so it won't exist if you just call the serialized interface (`dump`) and don't write other information before.

For the mge model format that can be visualized by netron, there is only one possible addition of information: mgbtest0. The added information is used to represent the presence of user-given input data in the model and can be identified by the tools (mge/load_and_run) that we provide to the user to analyze the inference performance of the model.

In general, model serialization should only be model serialization, but in mge, these short lead information were added to accommodate the use of other tools. So if it's more than 24 bytes, I think we should modify the way our tool recognizes it?

source/megengine.js
Ysllllll: The mapping here only involves some special data types (such as quantizeds4asymm), but there is no mapping for common data types (such as float32). For mge models in other formats (such as .tm format), the special data types they support are also included in this correspondence. So don't generalize? I wonder if I understand what you mean correctly?

tools/megengine
Ysllllll: ok

source/megengine.js
lutzroeder: Is `mgbtest0` a known header format? For example, detect, read its length, skip over it?

source/megengine.js
lutzroeder: Instead of mapping `quantizeds4asymm` to `uint8`, is `quantizeds4asymm` a common data type that is supported by other frameworks? What would be the right name to use for all frameworks? Should `dialog.Tensor` support these data types? For example, there is `int4' in other frameworks, if it is not encoded in a standard way maybe it should report the standard name and not return data for rendering?

source/megengine.js
Ysllllll: Yes, there will be a fixed 12-Byte offset. Fixed in this way.

source/megengine.js
Ysllllll: Data type `quantizeds4asymm` is defined by us for quantifying data types, so it is not an common data type supported by other frameworks. Its corresponding name in `dialog.Tensor` should be `quint8`, which is also the name used by other frameworks. done.

tools/megengine
Ysllllll: The problem is `sed -i 's/struct/table/g' ./third_party/source/megengine/src/serialization/fbs/*[^dtype].fbs` in `tools/megengine sync`. When using `tools/flatc.js` to generate `schema.js`, the deserialization method of `struct` should be generated instead of `table`.

When there is `struct`, I find that there is a problem when using `tools/flatc.js` to generate `schema.js`, as follows:
```javascript
// test.fbs
namespace mgb.serialization.fbs;

enum AxisDescMethod : byte {
    ADD_1,
    REMOVE,
}

struct AxisDesc {
    method:AxisDescMethod;  // change to method:byte;
    axis:int;
}

table AxisAddRemove {
    desc:[AxisDesc];
}

root_type AxisAddRemove;
```
For the `test.fbs` file above, `tools/flatc.js` can't work normally, However, if the `method:AxisDescMethod` field in `struct AxisDesc` is changed to `method:byte` (as shown in the following code block), `tools/flatc.js` can work normally. Is it because the generation cannot be nested? Can you help me to look at this problem?

```javascript
// test.fbs
namespace mgb.serialization.fbs;

enum AxisDescMethod : byte {
    ADD_1,
    REMOVE,
}

struct AxisDesc {
    method:byte;  // change from method:AxisDescMethod;
    axis:int;
}

table AxisAddRemove {
    desc:[AxisDesc];
}

root_type AxisAddRemove;
```

Ysllllll: When there is `struct`, I find that there is a problem when using `tools/flatc.js` to generate `schema.js`, as follows:
```javascript
// test.fbs
namespace mgb.serialization.fbs;

enum AxisDescMethod : byte {
    ADD_1,
    REMOVE,
}

struct AxisDesc {
    method:AxisDescMethod;  // change to method:byte;
    axis:int;
}

table AxisAddRemove {
    desc:[AxisDesc];
}

root_type AxisAddRemove;
```
For the `test.fbs` file above, `tools/flatc.js` can't work normally, However, if the `method:AxisDescMethod` field in `struct AxisDesc` is changed to `method:byte` (as shown in the following code block), `tools/flatc.js` can work normally. Is it because the generation cannot be nested? Can you please help me to look at this problem?

```javascript
// test.fbs
namespace mgb.serialization.fbs;

enum AxisDescMethod : byte {
    ADD_1,
    REMOVE,
}

struct AxisDesc {
    method:byte;  // change from method:AxisDescMethod;
    axis:int;
}

table AxisAddRemove {
    desc:[AxisDesc];
}

root_type AxisAddRemove;
```

Ysllllll:
Hi @lutzroeder, can this mr be merged?

lutzroeder:
@Ysllllll can you rebase to TIP. The `mgbtest0` and `./tools/megengine` changes should be no longer needed. If not, please share a small repro. The data type is still an open discussion. For now, use the specific name if there is no general name to get unblocked. This will result in the `uint8` not getting rendered which can be addressed in a later change.

Ysllllll:
OK. Yes, for `./tools/megengine`, its changes is no longer needed. for `mgbtest0`, need to modify a little bit, that is, remove the if statement of the file size(` if (size === (stream.length - position - 4))`), Here is an example:
[add_with_data.zip](https://github.com/lutzroeder/netron/files/10529117/add_with_data.zip)

PR #1057, opened 2023-03-10
Add support for ML.NET models with .mlnet extension

luisquintanilla:
ML.NET models generated by Model Builder now output files with the *.mlnet* extension. Everything about the file is the same. The only thing that's changed is the extension.

### DIFF
--- a/source/mlnet.js
+++ b/source/mlnet.js
@@ -7,7 +7,7 @@ var base = require('./base');
 mlnet.ModelFactory = class {
 
     match(context) {
-        const entries = context.entries('zip');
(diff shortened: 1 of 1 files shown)


luisquintanilla:
@lutzroeder I took a stab at adding support for ML.NET model files with the *.mlnet* extension. I'm not sure if this is right. I noticed the [tf.js](https://github.com/lutzroeder/netron/blob/main/source/tf.js) file for example does checks against the extension. 

In this case, nothing has changed about the way the file is serialized. The only thing that's changed is the extension. model files with the *.zip* extensions still work.

Issue #181, opened 2018-11-15
Show connections on hover


lutzroeder:

yxchng:
Just wondering, what is the status of this issue?

heyong919:
+1

nguyenvulong:
+1 for draggable objects.

PR #1071, opened 2023-04-15
show connections on hover

beru:
#181 

image


lutzroeder:
@beru can this be accomplished without doubling the number of edges in the SVG?

beru:
@lutzroeder Sure, we can use SVG `<use>` tag. I've done that in #1075 .

lutzroeder:
@beru this is not working in Safari on macOS and iPad (see #1106).

PR #1072, opened 2023-04-15
Use PointerEvent interface to enable pointer capturing

beru:
By using the `setPointerCapture()` method of the Element interface, graph scrolling would be more comfortable. It seems using PointerEvent is necessary for  the method to work so several event handlings are also updated.

I still haven't tested if the change is compatible with touch-screen devices. 

https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture
https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent

### DIFF
--- a/source/view.js
+++ b/source/view.js
@@ -344,9 +344,8 @@ view.View = class {
             this._events = {};
             this._events.scroll = (e) => this._scrollHandler(e);
             this._events.wheel = (e) => this._wheelHandler(e);
-            this._events.mousedown = (e) => this._mouseDownHandler(e);
             this._events.gesturestart = (e) => this._gestureStartHandler(e);
-            this._events.touchstart = (e) => this._touchStartHandler(e);
+            this._events.pointerdown = (e) => this._pointerDownHandler(e);
         }
         const graph = this._element('graph');
         if (graph) {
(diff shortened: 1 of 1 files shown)


lutzroeder:
@beru can you add some context which scrolling scenario, operating system, browser and device this is targeting?

beru:
@lutzroeder I use desktop app Netron, Version 6.8.2 on Windows 10. I scroll by clicking and dragging with a mouse.

Here's the scrolling scenario.

## Before

The scrolling is unexpectedly interrupted when a mouse cursor goes of a window during dragging.

before

## After

By using `setPointerCapture()`, the scrolling isn't interrupted when a mouse cursor goes of a window while dragging.

after

lutzroeder:
@beru does not work in Safari on macOS.

source/view.js
```diff
@@ -344,21 +344,19 @@ view.View = class {
...
+            this._events.pointerdown = (e) => this._pointerDownHandler(e);
         }
         const graph = this._element('graph');
         if (graph) {
             graph.focus();
         }
         graph.addEventListener('scroll', this._events.scroll);
         graph.addEventListener('wheel', this._events.wheel, { passive: false });
-        graph.addEventListener('mousedown', this._events.mousedown);
         if (this._host.environment('agent') === 'safari') {
```
beru, line 356: Due to this condition, Safari doesn't use pointerdown event.

The pointerdown event is supported from Safari 13.
https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerdown_event#browser_compatibility

Safari 13 was released on September 20, 2019
https://en.wikipedia.org/wiki/Safari_(web_browser)

lutzroeder:
@beru `e.touches` does not exist on `PointerEvent`, throws on iPad. Can you validate the commit at TIP.

beru:
@lutzroeder

> e.touches does not exist on PointerEvent, throws on iPad.

Oh I see... sorry for not testing on touch devices.

> Can you validate the commit at TIP.

I've tested b22093ccefdd5f97c161c8b7662225e7597bf669. It works well at my side.

PR #1078, opened 2023-04-18
Show connection on hover from view.FindSidebar

beru:
This is a sequel to #1075 .

hover


lutzroeder:
@beru the goal is to decouple `FindSidebar` from accessing the graph DOM tree directly, e.g. remove all dependencies on `this._graphElement` instead of adding new ones. The implementation is also accessing private members of the intermediate tree.

PR #1081, opened 2023-04-18
Show connection on hover without duplicating edges

beru:
Hit-testing is performed with SVG `<use>` tag so there's no need to duplicate edges anymore.

### DIFF
--- a/source/grapher.css
+++ b/source/grapher.css
@@ -64,13 +64,12 @@
 .graph-item-output:hover path { fill: #fff; }
 
 .edge-label { font-family: -apple-system, BlinkMacSystemFont, "Segoe WPC", "Segoe UI", "Ubuntu", "Droid Sans", sans-serif, "PingFang SC"; font-size: 10px; }
-.edge-path { stroke: #000; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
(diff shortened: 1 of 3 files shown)


lutzroeder:
@beru any insights which approach is more efficient in resource usage?

beru:
> any insights which approach is more efficient in resource usage?

I checked memory consumption of Electron processes with Task Manager.

It seems like this MR consumes a bit of more memory compared to main branch.

If lowering memory usage is important, I'd add a command-line option to stop adding the `hitTestElement` in `grapher.Edge.build` method. If browser users say Netron app has gotten resource hogging, adding the `hitTestElement` can be stopped by default.

lutzroeder:
@beru both memory and performance. For example, does it impact rendering speed of large graphs? The goal is to keep the app and browser version simple and easy to use while options tend to be hard to discover. It's more about understanding the impact of this feature and what potential issues might arise and the finding the best possible technical trade-off.

beru:
I guess it would impact rendering speed of large graphs that contain a lot of edges (let's say more than a few thousands). But I'd like to believe svg rendering of speed Chromium isn't that bad with recent GPUs.

Before rendering speed issue would arise, I suspect large graph layout with `dagre.js` won't finish so most of users wouldn't notice the rendering slowness. GraphViz and MSAGL are also pretty slow to layout when the number of nodes and edges exceed beyond 5,000 and I guess this is a nature of sugiyama algorithm.

PR #1084, opened 2023-04-21
Enable toggling Full Screen

beru:
#1085

### DIFF
--- a/source/app.js
+++ b/source/app.js
@@ -249,6 +249,7 @@ app.Application = class {
             case 'close': window.close(); break;
             case 'quit': electron.app.quit(); break;
             case 'reload': this._reload(); break;
+            case 'fullscreen': this._fullscreen(); break;
             case 'report-issue': electron.shell.openExternal('https://github.com/' + this._package.repository + '/issues/new'); break;
             case 'about': this._about(); break;
             default: {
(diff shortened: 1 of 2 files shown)


lutzroeder:
@beru what specific issue is this addressing and can it be accomplished without adding another menu item? 

For example, on macOS full screen is hidden in the top left menu:

image

beru:
> @beru what specific issue is this addressing and can it be accomplished without adding another menu item?

I'vce opened #1085 to accommodate the request.

Without adding another menu item would be possible. But to do so, there needs to be a way to go back from fullscreen without using a keyboard. Otherwise, some users might complain they had to restart/shutdown their computers to exit from Netron.

beru:
@lutzroeder I've just installed Version 6.8.4 and checked Full Screen toggling.

-----

Before pressing F11 key.

image

After pressing F11 key, the window is maximized and switched to Full Screen but Minimize/Mazimize/Close icons are still shown. They shouldn't appear when the window is in Full Screen.

image

Pressing F11 key again to go back from Full Screen. Oddly, Minimize/Mazimize/Close icons are hidden.

image

---------

To prevent the above weird behavour, please check what I did in #1084 

https://github.com/beru/netron/blob/9c730f0e13ba6a70911fd5c6f4a2bba5296b713c/source/app.js#L822-L835

beru:
I've confirmed that the problem has fone with the recent commit (c4f2cea932af1a28028af01fbfb7912117d595fd), thank you.

beru:
@lutzroeder While Netron is running, Other applications such as VSCode, Microsoft Edge don't react to F11 key for toggling Full Screen.

Issue #1125, opened 2023-07-17
Toggle visibility of tensor sizes (shapes) on edge labels


beru:
Currently, sizes of tensors (shapes) aren't displayed when names are shown. But in some situations, I'd like to see both of edge names and sizes altogether.

PR #1126, opened 2023-07-17
Toggle visibility of sizes (tensor shapes)

beru:
This PR adds `Show/Hide Sizes` menu-item to toggle visibility of sizes on edge labels.

toggle_sizes

#1125


lutzroeder (changes requested):
* To reduce the number of options, can "Show Names" show both the name and shape (if available).
* The labels are not vertically centered when both name and shape is shown.

screenshot

PR #1158, opened 2023-09-28
Add example of OpenVINO IR to open by Netron

rkazants:


rkazants:
@lutzroeder, could you please review?

Best regards,
Roman

lutzroeder:
The goal is to keep this samples list short. Is there a reason or data OpenVINO should be prioritized over the other formats?

rkazants:
> @rkazants the goal is to keep the samples list short.
> 
> Is there a reason or data OpenVINO should be prioritized over the other formats?

Hi @lutzroeder,

I see your point. I thought the goal of this list is to highlight that Netron is very versatile and supports variety of FWs formats from different companies proven by the concrete examples. You can optimize the current list by combining TensorFlow and TensorFlow Lite examples because it looks like repetitions in the list.

Thanks a lot,
Roman

Issue #1151, opened 2023-09-14
Keras: Functional support


Chizkiyahu:
- Netron app and version:  website 
- OS and browser version: Mac OS 13.5.2 Chrome 116.0.5845.187

Steps to Reproduce:

1. create a tf model and use the same layer object multiple times
-  multiple inputs to the same object (model1)
-  output of layer goes to the same layer (model2)
-  output of layer goes to the same layer with another layer in the middle (model3)

```
import tensorflow as tf

keras_input1 = tf.keras.Input((3, 10, 10))
keras_input2 = tf.keras.Input((3, 10, 10))
keras_input3 = tf.keras.Input((3, 10, 10))
tf_conv = tf.keras.layers.Conv2D(filters=16, kernel_size=3)
out1 = tf_conv(keras_input1)
out2 = tf_conv(keras_input2)
out3 = tf_conv(keras_input3)
inputs = [keras_input1, keras_input2, keras_input3]
outputs = [out1, out2, out3]

model1 = tf.keras.Model(inputs=inputs, outputs=outputs)
model1.save("model1.keras")

keras_input = tf.keras.Input((3, 10, 10))
tf_conv = tf.keras.layers.Conv2D(filters=10, kernel_size=3, padding="same")
out = tf_conv(keras_input)
for _ in range(5):
    out = tf_conv(out)
model2 = tf.keras.Model(inputs=[keras_input], outputs=[out])
model2.save("model2.keras")

keras_input = tf.keras.Input((3, 10, 10))
tf_conv = tf.keras.layers.Conv2D(filters=10, kernel_size=3, padding="same")
out = tf_conv(keras_input)
for _ in range(5):
    out = tf.keras.layers.ReLU()(out)
    out = tf_conv(out)
model3 = tf.keras.Model(inputs=[keras_input], outputs=[out])
model3.save("model3.keras")

```

2. open in netron 
- inputs name are not true -  is showing [ input kernel bias 0 1] and is [ input1, input2, input3, kernel, bias]
- output is showing only 1 
- the edges are okay in model1 and model3 but in model2 it looks like only goes from node to the same node but is actually 5-edge 
![Screenshot 2023-09-14 at 13 47 11](https://github.com/lutzroeder/netron/assets/37312901/916c74d5-935d-4c0d-867b-c194fcfb60c2)
[...]

PR #1159, opened 2023-10-06
keras fix layer sharing inputs names

Chizkiyahu:
fix the first issue in  #1151


Chizkiyahu:
I found more edge cases and working to fix them 
I will reopen when is ready

PR #1163, opened 2023-10-07
keras fix layer sharing inputs and outputs names

Chizkiyahu:
fix layer-sharing inputs and outputs #1151 
before fix 
Screenshot 2023-10-07 at 21 35 28
after fix 

Screenshot 2023-10-07 at 21 34 51

### DIFF
--- a/source/keras.js
+++ b/source/keras.js
@@ -788,21 +788,54 @@ keras.Node = class {
 
         const innerType = this.inner ? this.inner.type : null;
         const innerMetadata = innerType ? metadata.type(innerType) : null;
+
+        // handle layer sharing
         let inputIndex = 0;
+        let inputTypes = [];
+        const outputTypes = [];
+        if (this._type && this._type.inputs) {
+            for (let i = 0; i < this._type.inputs.length; i++) {
+                if (layer.inbound_nodes && layer.inbound_nodes.length > 1 && layer.inbound_nodes[0].length > i) {
+                    for (let j = 0; j < layer.inbound_nodes.length; j++) {
+                        const inputType = JSON.parse(JSON.stringify(this._type.inputs[i]));
+                        inputType.name =  j + ": "+ inputType.name;
+                        inputTypes.push(inputType);
+                    }
+                } else {
+                    inputTypes.push(this._type.inputs[i]);
+                }
+            }
+        }
+        if (this._type && this._type.outputs) {
+            for (let i = 0; i < this._type.outputs.length; i++) {
+                if (layer.inbound_nodes && layer.inbound_nodes.length > 1 && layer.inbound_nodes[0].length > i) {
+                    for (let j = 0; j < layer.inbound_nodes.length; j++) {
+                        const outputType = JSON.parse(JSON.stringify(this._type.outputs[i]));
+                        outputType.name = j + ": "+ outputType.name;
+                        outputTypes.push(outputType);
+                    }
+                } else {
+                    outputTypes.push(this._type.inputs[i]);
+                }
+            }
+        }
+        const inbound_nodes_size = layer.inbound_nodes ? layer.inbound_nodes.length : 1;
+        outputs = Array(inbound_nodes_size).fill(outputs).flat();
+
         while (inputs.length > 0) {
             let list = false;
             let name = null;
             let visible = true;
             if (!innerMetadata || inputIndex == 0) {
-                if (this._type && this._type.inputs && inputIndex < this._type.inputs.length) {
-                    const input = this._type.inputs[inputIndex];
+                if (inputTypes && inputIndex < inputTypes.length) {
+                    const input = inputTypes[inputIndex];
                     name = input.name;
                     if (type === 'BatchNormalization' && name === 'gamma' && config.scale === false) {
                         inputIndex++;
                         continue;
                     }
                     visible = input.visible == false ? false : true;
-                    if (this._type.inputs[inputIndex].list) {
+                    if (inputTypes[inputIndex].list) {
                         list = true;
                     }
                 }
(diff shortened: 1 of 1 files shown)


lutzroeder:
@Chizkiyahu double checking, this is a fix for the 1st screenshot in #1151, not the 2nd and 3rd?

Chizkiyahu:
> @Chizkiyahu double checking, this is a fix for the 1st screenshot in #1151, not the 2nd and 3rd?

this fixes the 1st and 3nd 
on the 2nd  I am undecided because when you click on the edge you see the current data

lutzroeder:
@Chizkiyahu is this the right approach? For Keras Functional there seems to be a differentiation between nodes and layers. The current graph is built from layers instead of node instances. For example, #3 should show multiple node instances of the Convolution layer. e.g. is the problem making the graph look more right or is it more that the layers don't match the underlying Keras Functional representation?

Chizkiyahu:
@lutzroeder 
In Keras, a layer can have more than one node 

| Comparison Criteria            | Nodes                                        | Layers                                       |
|--------------------------------|----------------------------------------------|----------------------------------------------|
| Graph Clarity                  | ✅ More accurate and clear                    | ❌ Can become less clear with complexity     |
| Interaction Requirement        | ✅ Clear without clicking                     | ❌ May require interaction to understand fully |
| Shape Inference                | ✅ Can show shape inference                   | ❌ Shape inference may not always be possible  |
| Complexity                     | ❌ Represents more complex graphs             | ✅ Simplifies to high-level abstractions     |
| Readability                    | ❌ Less readable as complexity increases      | ✅ More readable due to abstraction          |
| Data Duplication               | ❌ May duplicate data like constants and attributes | ✅ Tends to minimize data duplication |
| Netron Implementation          | ❌ Needs to implement                         | ✅ Already implemented in Netron             |
| Similarity to User Code        | ❌ May differ significantly                   | ✅ Closely resembles user-written code       |

My personal view is that there is no clear advantage to moving away from the existing 'layers' mode to a 'nodes' mode. It would be beneficial to keep the visualization focused on layers and simply indicate places where multiple nodes are present.

## Internal Project: Model Flattening

### Overview
I'm engaged in an internal project that focuses on flattening complex neural network models. This process includes challenging scenarios such as embedding a model within another model ('model-in-model') and the sharing of layers between different parts of the network.

### Demonstrating Complexity
To highlight the intricate nature of flattening such models, I've created a test and provided examples of both the original and the flattened models:

- Original model: [model.keras.zip](https://github.com/lutzroeder/netron/files/13257201/model.keras.zip)
- Flattened model: [flat_model.pb.zip](https://github.com/lutzroeder/netron/files/13257202/flat_model.pb.zip)

These examples represent an extreme case, showcasing the increase in complexity when flattening models. It is important to note that the visualization tool Netron does not inherently flatten 'model-in-model' architectures.

### Conclusion
The provided examples underscore the sophisticated nature of model architectures and the potential complications that arise from attempting to flatten such structures for visualization or analysis.

lutzroeder:
@Chizkiyahu yes, the table is clear.

* Is it possible to implement this without relying on information from `keras-metadata.json`? For example, summarizing how many inputs are used shouldn't depend on looking at `type.inputs`?
* Is this about flattening graphs? In the #3 example the output seems incorrect as the order in which `ReLU` is called is lost. `Conv2D` seems to be more a function module that is defined once and invoked via `conv2d[0][0]`, `conv2d[1][0]`, conv2d[2][0]` instance nodes?

Chizkiyahu:
> @Chizkiyahu yes, the table is clear.
> 
> * Is it possible to implement this without relying on information from `keras-metadata.json`? For example, summarizing how many inputs are used shouldn't depend on looking at `type.inputs`?
> * The question was less about flattening graphs. In the 2) and 3) examples, the output seems incorrect as the order in which `ReLU` is called is lost. `Conv2D` seems to be more a function module that is defined once and invoked via `conv2d[0][0]`, `conv2d[1][0]`, conv2d[2][0]` instance nodes? e.g. what is the exact meaning of a layer and node in Keras.

# Keras Nodes vs Layers

In Keras, nodes represent the points within a network graph where layers can connect, whereas a layer is an instance of a Keras layer that performs a specific function within the neural network. It is possible to add multiple nodes to a network using the same instance of a Keras layer. This concept is represented in a Keras model as a single layer with multiple nodes.

Graph of Keras nodes is [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) but the graph of keras nodes can be a `cyclic graph`

# Input and Output Order

The order of edges can be made clearer by numbering them instead of using "to" This can be observed when selecting the output edge of a conv layer.

Diagram Illustration

# Implementation Methods

- **Graph of Keras Layers (Main):** Focuses solely on layers, which have the bugs.
- **Graph of Keras Nodes:** Flattens the graph.
- **Graph of Keras Layers and considers/combines the Keras Nodes (PR):**  graph of layers that show all the cyclic inputs numbered 

Please tell me if you have another Idea of how to implant this 

# Possible Improvements

- Add a numbered indicator on the Netron node (e.g., `Conv2D`) to reflect the number of Keras nodes (`layer.inbound_nodes.length`).
- Number the edges for clear identification.
- Add a 'Keras Nodes Mode'/`DAG` in settings for alternative visualization.

lutzroeder:
@Chizkiyahu [main](https://github.com/lutzroeder/netron) now has a `node_mode` (disabled by default) to experiment with flattening the graph. For [model.keras.zip](https://github.com/lutzroeder/netron/files/13257201/model.keras.zip) shared above, does the model have dependencies across multiple layers? For example, there is a reference to `model_3[4]` and `model_3` has only 4 inbound nodes.

image

lutzroeder:
@Chizkiyahu [main](https://github.com/lutzroeder/netron) now has a `node_mode` experiment to partially inline layers with multiple inbound nodes. Layers with multiple nodes get lifted as functions. 

image

PR #1173, opened 2023-10-19
OpenVINO: fix i4/u8 itemsize

yli150:
I4/U4 does not have right itemSize set which leads to the raising of error.  This fix is to set 0.5 for I4/U4 precision.

[openvino_i4_matmul_model.zip](https://github.com/lutzroeder/netron/files/13052433/openvino_i4_matmul_model.zip)


lutzroeder:
@yli150 can you please share a sample file to validate the change or add a test case to `./test/models.json`.

yli150:
I have added it into model test list.  Please help to review.

PR #1174, opened 2023-10-19
Add attribute-search to the left sidebar

ijprins:
This makes it a lot easier to work with graphs where nodes have a large number of attributes, only a few of which are important.
See the attached screenshot for an example of it working an onnx graph (not really an example of the use-case however).
I didn't realize that electron doesn't have the left sidebar until pretty late, I'm not sure you can directly transpose the code into that version either since I don't *think* the OS top bar will let you add a text-input control, but I'm no expert.
I figured I'd let you decide what's best to do there.
Lint is clean and tests seem to be fine, though I'm not sure that they exercise this code by nature.

filter-attributes-netron


lutzroeder:
@ijprins as you pointed out this doesn't work on all platforms. The left sidebar is a menu and this is rather specific feature that probably should be in a more hidden location. Is typing property names and pressing enter the right approach for touch screens or are there other ways to enable this? Stepping back a little, is there evidence that this is a common problem that needs solving given there isn't an upvoted issue? Is it more about filtering or more about finding? For example, would a special option in the Find dialog that highlights attributes or weights be more broadly applicable?

PR #1186, opened 2023-11-17
Adding Modular support to Netron

katelyncaldwell:
Adding the initial support for Modular visualization files (`.mod`) to Netron. This is the first version of this support, and we expect to expand it to add support for shapes and additional attributes in later iterations. 

More context on modular can be found at [modular.com](https://www.modular.com/). Additional information about how to utilize this integration will be released in a few weeks.


katelyncaldwell:
Modular test file: 
[resnet34.mod.zip](https://github.com/lutzroeder/netron/files/13395132/resnet34.mod.zip)

PR #1191, opened 2023-11-23
Improve SVG building performance

AkiSakurai:
The getBBox function is slow. This commit replaces it with measureText to enhance performance.

Before
Screenshot 2023-11-23 at 11 03 09 AM

After:
Screenshot 2023-11-23 at 11 04 37 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
         "electron": "27.1.0",
         "electron-builder": "24.6.4",
         "@electron/notarize": "2.2.0",
-        "eslint": "8.54.0"
+        "eslint": "8.54.0",
+        "canvas": "^2.11.2"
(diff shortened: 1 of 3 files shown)


lutzroeder (changes requested):
@AkiSakurai can you provide some context:
1) What are steps to reproduce and why is this an issue?
2) Has the change been tested for all platforms, browser versions, display settings?
3) Are the results exactly the same?

AkiSakurai:
1. Opening a graph with more than 1000 nodes takes a long time.
3. The result is not exactly the same; there are subpixel differences.

2. The fontBoundingBoxDescent property is not available in Firefox, so it will not work in Firefox.

PR #1238, opened 2024-03-01
Update tflite.js

Rob-Hughes-Arm:
**Support input/output names from signature_defs**

.tflite files can have additional data called SignatureDefs which provide more descriptive names for the network inputs and outputs (e.g. "image_in" rather than "Idenity 177"). This patch makes it so that this data (when present) is used to set the name of the input and output tensors so that networks are easier to read.

The schema is here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/schema/schema.fbs#L1592

### DIFF
--- a/source/tflite.js
+++ b/source/tflite.js
@@ -1,3 +1,4 @@
+// SPDX-FileCopyrightText: Copyright 2024 Arm Limited and/or its affiliates open-source-office@arm.com
 
 import * as flatbuffers from './flatbuffers.js';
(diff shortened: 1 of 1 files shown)


lutzroeder:
@Rob-Hughes-Arm please add a small test file to `./test/models.json` and run `npm run validate`.

Rob-Hughes-Arm:
I see you merged this change into another patch - thanks! :D

PR #1244, opened 2024-03-15
Make dagre run faster with runLayout

seanshpark:
This is a small hack to make dagre runLayout method run faster.

### DIFF
--- a/source/dagre.js
+++ b/source/dagre.js
@@ -887,8 +887,10 @@ dagre.layout = (graph, layout) => {
                 }
             };
             const queue = g.children();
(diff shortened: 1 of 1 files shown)


seanshpark:
It's a hack so I'm not sure you can accept this but
this small changes made my LLM loading time from 15 minutes to about 2 minutes 30 seconds.

seanshpark:
FYI issue about Array.shift in chromium is https://bugs.chromium.org/p/v8/issues/detail?id=12730

seanshpark:
Thanks for accepting!

Issue #864, opened 2022-01-06
Graph visualization unnecessarily wide


leosh64:
For some networks, the visualization of the graph becomes unnecessarily wide. This makes it difficult to understand connections when zoomed in:

![image](https://user-images.githubusercontent.com/4062160/148372676-db4ca6d6-a757-4fe0-a59a-dbec10e46cd2.png)

Throughout the model, there is no section that would justify such a wide visualization.

Below some screenshots when zoomed out:

![image](https://user-images.githubusercontent.com/4062160/148372478-c0746880-0d4d-4045-885a-95fd021325cd.png)

The same issue (unnecessarily high) occurs in horizontal mode:

![image](https://user-images.githubusercontent.com/4062160/148372584-1318f6a3-1777-47e4-a670-41f60ae8d3c6.png)

KarelPeeters:
I was just about to report this as well!

For example, [yolov5n.onnx](https://github.com/ultralytics/yolov5/releases/download/v6.0/yolov5n.onnx) looks like this for me:

![image](https://user-images.githubusercontent.com/7887115/148389610-ff5d9746-7254-441f-b021-c51bea355666.png)

This width is somewhat justifiable since the final layers of the model do indeed need the width, but there is no reason for the connections that go from a conv to the following mul to go all the way to the right and back.

lutzroeder:
@leosh64 @KarelPeeters can you help investigate this and suggest a fix.

dagrejs/dagre#239 has some context. The paper describing the algorithm can be found [here](https://www.semanticscholar.org/paper/Fast-and-Simple-Horizontal-Coordinate-Assignment-Brandes-Köpf/69cb129a8963b21775d6382d15b0b447b01eb1f8) and the code [here](https://github.com/lutzroeder/netron/blob/1c8316d6a99111eee8fbe91efb85913a43d5e791/source/dagre.js#L1806-L1843).

lutzroeder:
[yolov5n.onnx.zip](https://github.com/lutzroeder/netron/files/7962856/yolov5n.onnx.zip)

Fannettembrown:
Is your phone build with special lenses?

AlessandroMondin:
Hello all, any news on this bug?
I am facing the same issue as weel: 
<img width="998" alt="Screenshot 2022-08-31 at 12 56 39" src="https://user-images.githubusercontent.com/80039909/187663428-464a3655-6f54-4883-b293-b689b24ae6fa.png">

[...]

PR #1246, opened 2024-03-18
Fix dagre.js for abnormal wide edges

seanshpark:
This will dagre.js for abnormal wide edges

### DIFF
--- a/source/dagre.js
+++ b/source/dagre.js
@@ -1591,7 +1591,7 @@ dagre.layout = (graph, layout) => {
                             }
                             const label = g.node(v).label;
                             if (min !== Number.POSITIVE_INFINITY && label.borderType !== borderType) {
-                                xs[v] = Math.max(xs[v], min);
+                                xs[v] = Math.min(xs[v], min);
                             }
                         } else {
                             visited.add(v);
@@ -1740,7 +1740,7 @@ dagre.layout = (graph, layout) => {
                 for (const [v, x] of Object.entries(xs)) {
                     const halfWidth = g.node(v).label.width / 2;
                     max = Math.max(x + halfWidth, max);
-                    min = Math.min(x - halfWidth, min);
+                    min = Math.min(Math.max(0, x - halfWidth), min);
                 }
                 const width = max - min;
                 if (width < minWidth) {


seanshpark:
this is to fix #864 issue

source/dagre.js
```diff
@@ -1591,7 +1591,7 @@ dagre.layout = (graph, layout) => {
                             }
                             const label = g.node(v).label;
                             if (min !== Number.POSITIVE_INFINITY && label.borderType !== borderType) {
-                                xs[v] = Math.max(xs[v], min);
+                                xs[v] = Math.min(xs[v], min);
```
seanshpark, line 1594: this will choose smallest width

source/dagre.js
```diff
@@ -1740,7 +1740,7 @@ dagre.layout = (graph, layout) => {
                 for (const [v, x] of Object.entries(xs)) {
                     const halfWidth = g.node(v).label.width / 2;
                     max = Math.max(x + halfWidth, max);
-                    min = Math.min(x - halfWidth, min);
+                    min = Math.min(Math.max(0, x - halfWidth), min);
```
seanshpark, line 1743: some `x - halfWidth` goes to minus, which will make below `const width = max - min;` larger.

lutzroeder:
This is changing how other models are rendered? [sample.zip](https://github.com/tensorflow/tensorflow/files/917065/sample.zip)

Before:



After:

seanshpark:
> Is there a good way to reason pros vs. cons?

@ lutzroeder, thanks for looking into this !
I'll take a look at the sample model :)

seanshpark:
@lutzroeder , as not knowing much about algorithms in dagre, my wild guess is that the end node of the example graph tries to go to the center of the layout (the minimum diff from center?)

I'll try to study what the algorithm does and find a way for this.
If the example is important, you can just ignore this PR,
or if most of the well know models work, I would be glad if you can land this change.

Anyway, I'll try to find something better.

seanshpark:
Close this for now and reopen when change is ready.

source/dagre.js
beru: @seanshpark @lutzroeder 

How about the following code?

```js
                            let minOut = Number.POSITIVE_INFINITY;
                            let minIn = Number.POSITIVE_INFINITY;
                            for (const e of blockG.node(v).out) {
                                minOut = Math.min(minOut, xs[e.w] - e.label);
                            }
                            for (const e of blockG.node(v).in) {
                                minIn = Math.min(minIn, xs[e.w] - e.label);
                            }
                            const label = g.node(v).label;
                            if (minOut !== Number.POSITIVE_INFINITY && minIn !== Number.NEGATIVE_INFINITY && label.borderType !== borderType) {
                                if (Math.abs(xs[v] - minOut) < Math.abs(xs[v] - minIn)) {
                                    xs[v] = Math.max(xs[v], minOut);
                                }else {
                                    xs[v] = Math.max(xs[v], minIn);
                                }
                            }
```

model | before | after
------|--------|-------
yolov5n | yolov5n onnx | new_yolov5n onnx
sample.zip | saved_model pbtxt | new_saved_model pbtxt

seanshpark: > How about the following code?

This looks better !

seanshpark:
I couldn't repoen this so posted another  #1254

PR #1254, opened 2024-04-01
Fix dagre.js for abnormal wide edges

seanshpark:
This will update dagre.js for abnormal wide edges

### DIFF
--- a/source/dagre.js
+++ b/source/dagre.js
@@ -1647,6 +1647,9 @@ dagre.layout = (nodes, edges, layout, state) => {
                             min = Math.min(min, xs[e.w] - e.label);
                         }
                         const label = g.node(v).label;
+                        if (label.dummy) {
+                            continue;
+                        }
                         if (min !== Number.POSITIVE_INFINITY && label.borderType !== borderType) {
                             xs[v] = Math.max(xs[v], min);
                         }


seanshpark:
this change has suggestion from @beru at https://github.com/lutzroeder/netron/pull/1246#discussion_r1546742639

seanshpark:
I'm still reading the papers of Dagre ... and looks like will take a while.

seanshpark:
@beru, Wow, you've got the sample working !

> It was difficult for me to understand `buildBlockGraph` and the second pass of `horizontalCompaction` 

Algorithms are described in the papers in https://github.com/dagrejs/dagre/wiki#recommended-reading (I think you may know the links) and still reading these hard-to-understand algorithms in my spare time ... 
I'll also take a look at your code :)
BTW, what about posting a PR ?

beru:
> Algorithms are described in the papers in https://github.com/dagrejs/dagre/wiki#recommended-reading (I think you may know the links)

Thank you for the info. I didn't know the wiki of dagre project or maybe I just forgot it? Anyway, I'll try to understand what's written on [the paper](https://link.springer.com/content/pdf/10.1007/3-540-45848-4_3.pdf).

> BTW, what about posting a PR ?

I wish to do so if I'll be able to come up with more polished code. The code I wrote doesn't use the second pass at all (as I'm not capable of understanding it) but that's probablly not a good idea.

seanshpark:
@lutzroeder , I've update the patch with what I understand as of now.
For the problem of undesired long edges are from the dummy nodes in dagre.
This change will skip for calculating the position for dummy elements,
so that they don't go to some unwanted position.

But still there exist a problem with the model you've provided and the result is same as in https://github.com/lutzroeder/netron/pull/1254#issuecomment-2030692952 capture.
What I know is that this particular node input/output is reversed in Y position.
I hope I could find the cause to fix the degradation someday :)

If possible can you please land current change?

seanshpark:
@lutzroeder , I've updated for CI build fail.

PR #1260, opened 2024-04-11
MegEngine: fix duplicate value error

YSF-A:
fix duplicate value error of megengine  visualization

please help review, thanks~

the sample models:
[mge_duplicate_value.zip](https://github.com/lutzroeder/netron/files/14955823/mge_duplicate_value.zip)


lutzroeder:
@YSF-A please share a sample model file to validate this change.

PR #1269, opened 2024-04-25
Add icons to FIND side bar to toggle search item types

beru:
This PR adds icons to FIND side bar to toggle search item types (Edges/Nodes/Initializers).

Toggling off all of the icons is prevented.

### DIFF
--- a/source/index.html
+++ b/source/index.html
@@ -212,7 +212,11 @@ button { font-family: -apple-system, BlinkMacSystemFont, "Segoe WPC", "Segoe UI"
     background-repeat: no-repeat;
 }
 .sidebar-separator { margin-bottom: 20px; }
-.sidebar-find-search { font-family: inherit; font-size: 13px; margin: 0px 20px 8px 20px; padding: 8px 16px 8px 16px; background: #fff; border-radius: 16px; border: 0; outline: 0; }
+.sidebar-find-box { position:relative; }
+.sidebar-find-search { position:relative; width:85%; font-family: inherit; font-size: 13px; margin: 0px 20px 8px 20px; padding: 8px 16px 8px 16px; background: #fff; border-radius: 16px; border: 0; outline: 0; }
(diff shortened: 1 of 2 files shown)


source/index.html
```diff
@@ -212,7 +212,11 @@
     background-repeat: no-repeat;
 }
 .sidebar-separator { margin-bottom: 20px; }
-.sidebar-find-search { font-family: inherit; font-size: 13px; margin: 0px 20px 8px 20px; padding: 8px 16px 8px 16px; background: #fff; border-radius: 16px; border: 0; outline: 0; }
+.sidebar-find-box { position:relative; }
+.sidebar-find-search { position:relative; width:85%; font-family: inherit; font-size: 13px; margin: 0px 20px 8px 20px; padding: 8px 16px 8px 16px; background: #fff; border-radius: 16px; border: 0; outline: 0; }
+.siderbar-find-edge-icon { font-family: 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace; font-size: 12px; position:absolute; background-color:#0000; top:8px; right:40px; user-select: none; -webkit-user-select: none; -moz-user-select: none; }
```
lutzroeder, line 217: Instead of positioning manually using `top`, is there a way to center automatically? Is there a set of characters that is guaranteed to be aligned is all system fonts or should this change to use SVG icons? 

image

source/view.js
```diff
@@ -3253,6 +3253,53 @@ view.FindSidebar = class extends view.Control {
...
+            if (toggleInitializer) {
+                showInitializer = !showInitializer;
+            }
+            if (!showEdge && !showNode && !showInitializer) {
+                return;
+            }
+            this._showEdge = showEdge;
+            this._showNode = showNode;
+            this._showInitializer = showInitializer;
+            edgeIcon.style.opacity = this._showEdge ? "1.0" : "0.4";
```
lutzroeder, line 3285: Use CSS class instead of manually editing style properties?

lutzroeder (changes requested):

source/index.html
beru: I'm not really sure to be honest. The world of fonts looks really complicated. The unaligned positioning of the icons becomes rather obvious when the font-size gets bigger, so just as you mentioned, it's undesirable to position the icons manually.

image

Issue #1240, opened 2024-03-06
Metrics metadata support


lutzroeder:
**Scenarios:**

1) Allowing the app to contain implementations to compute metrics.  See [#204](https://github.com/lutzroeder/netron/issues/204).
2) Open existing files defined by framework and format vendors. Which formats and tools exist?
3) Supporting scripted solutions and formats from custom tools. See [#1234](https://github.com/lutzroeder/netron/issues/1234).

**Questions:**

- Are computed or loaded metrics different from metadata provided via [onnx/onnx#5938](https://github.com/onnx/onnx/pullrequests/5938) or [metadata_schema.fbs](https://github.com/tensorflow/tflite-support/tree/master/tensorflow_lite_support/metadata)?
- Should these values show as part of the Node metadata UI or separate?
- What API structure supports all scenarios?

kylesayrs:
I'd be interested in helping with this feature. I implemented a similar feature for Neural Magic's Sparsezoo [calculate_ops.py](https://github.com/neuralmagic/sparsezoo/blob/main/src/sparsezoo/utils/calculate_ops.py)

I propose support of both weight sparsity metrics and operations metrics.

Counting operations depends upon whether the runtime engine and hardware supports sparsity, block sparsity, and quantization. The UI design should be capable of supporting these subtypes, if not now then in the future.

1. Since [onnx/onnx#5938](https://github.com/onnx/onnx/pull/5938) and [metadata_schema.fbs](https://github.com/tensorflow/tflite-support/tree/master/tensorflow_lite_support/metadata) seem to be unstructured, supporting these kinds of visualizations seems to be a separate issue.
2. I propose showing weight sparsity within the node metadata UI alongside the weight, ie
```
name: model.3.conv.weight
category: Initializer
tensor: float32[64,32,3,3] (81% sparsity)
```

As for visualizing operations, I'm in favor of separating the UI from the node metadata tab as to make it clear that these performance (operation) metrics are computed values separate from the data which is embedded in the model file. For example, these could be a togglable UI displayed to the right of a node.

kylesayrs:
Another UI idea might be togglable horizontal bars which appear to the left of a node and have different sizes depending on how many operations of which types are associated with it.

There should also be a UI element for the total number of ops/sparsity in the model, perhaps in the bottom right

lutzroeder:
@kylesayrs all great questions.

Fundamentally, there seem to be 3 types of data and a question how these are unified and exposed at which API layer.

[...]

Issue #65, opened 2018-01-16
Show 1D/2D tensors as graph or color plot


lutzroeder:

goelbenj:
@lutzroeder is this still a desirable feature? I find myself needing this more than often.

If so, what types of visualizations do you feel is conducive to meaningful analysis?

Issue #1241, opened 2024-03-06
Graph annotation based on attribute values


lutzroeder:
**Scenario:**

UI feature to augment the graph with a visualization of a selected attribute, metadata or metrics values.

The selected attribute should both be shows as part of each node and if float turned into a border or background style.

**Questions:**

- Should the feature get activated from property value by selecting a metrics of by filtering (Find) based on criteria?
- What visualization is most intuitive?

SStarver:
Mark,  it would be fantastic to support this feature.  Recently, I'm looking for a method to fill accuracy analysing result into each onnx model tensor, I think this feature is useful to satisfy my demandings.

sheny1xuan:
Looking forward to this feature.

xieofxie:
Looking forward to this! For example, I want to debug QDQ to know which node significantly goes wrong, I could use this https://github.com/microsoft/onnxruntime-inference-examples/blob/main/quantization/image_classification/cpu/run_qdq_debug.py to generate a json file with name and accuracy, then show it side by side with graph in the netron

justinchuby:
It would be nice to have the ability to load additional data provider files for visualization. This way the data does not have to be part of the model file. Similar features were implemented in model explorer: https://github.com/justinchuby/model-explorer-onnx#visualizing-pytorch-onnx-exporter-dynamotrue-accuracy-results

PR #1293, opened 2024-06-13
Tensor sparsity metric

kylesayrs:
Implements sparsity metric for initializer tensors as a first step to supporting #1240 and #65.
* Add `style` member to `view.Argument` to store style
* Implement the `percentage` style for displaying percentages such as sparsity
* Add a metrics section to the `TensorSidebar` which displays tensor metrics
* Lazily calculate metrics of a tensor only when demanded by `TensorSidebar`

Tensor Properties sidebar now includes a metrics section with sparsity
Screenshot 2024-06-13 at 00 32 36

Tensors which are not initializers do not have metrics
Screenshot 2024-06-13 at 00 33 04

### DIFF
--- a/source/view.js
+++ b/source/view.js
@@ -1780,6 +1780,8 @@ view.Graph = class extends grapher.Graph {
     }
 
     add(graph, signature) {
+        this.identifier = this.model.identifier;
+        this.identifier += graph && graph.name ? `.${graph.name.replace(/\/|\\/, '.')}` : '';
         const clusters = new Set();
         const clusterParentMap = new Map();
         const groups = graph.groups;
--- a/source/view.js
+++ b/source/view.js
@@ -2687,7 +2687,7 @@ view.ValueTextView = class extends view.Control {
                         line.style.whiteSpace = style;
                         break;
                     case 'percentage':
-                        line.innerText = `${(item * 100).toFixed(3)}%`;
+                        line.innerText = `${(item * 100).toFixed(1)}%`;
                         break;
                     default:
                         line.innerText = item;
@@ -3167,7 +3167,7 @@ view.TensorSidebar = class extends view.ObjectSidebar {
             this.add('type', item);
         }
 
-        if (value.initializer && value.initializer.category == 'Initializer') {
+        if (value.initializer && value.initializer.category === 'Initializer') {
             this.addHeader('Metrics');
 
             const tensor = new view.Tensor(value.initializer);
@@ -3639,7 +3639,7 @@ view.Argument = class {
         this.name = name;
         this.value = value;
         this.type = type;
-        this.style = style
+        this.style = style;
     }
 };
 
@@ -4139,7 +4139,7 @@ view.Tensor = class {
         if (!keys.has('sparsity')) {
             let num_zeros = 0;
             let num_parameters = 0;
-            let stack = [value];
+            const stack = [value];
             while (stack.length > 0) {
                 const val = stack.pop();
                 if (Array.isArray(val)) {
@@ -4147,7 +4147,7 @@ view.Tensor = class {
                         stack.push(element);
                     }
                 } else {
-                    num_zeros += Number(val == 0);
+                    num_zeros += Number(val === 0);
                     num_parameters += 1;
                 }
             }
--- a/source/view.js
+++ b/source/view.js
@@ -3170,6 +3170,8 @@ view.TensorSidebar = class extends view.ObjectSidebar {
         if (value.initializer && value.initializer.category === 'Initializer') {
             this.addHeader('Metrics');
 
+            // view.Tensor is also initialized in view.ValueView
+            // future work could cache the first initialization 
             const tensor = new view.Tensor(value.initializer);
             const metrics = tensor.metrics;
             for (const metric of metrics) {
--- a/source/view.js
+++ b/source/view.js
@@ -3651,6 +3651,7 @@ view.Tensor = class {
         this._encoding = tensor.encoding;
         this._layout = tensor.type.layout;
         this._stride = tensor.stride;
+        this._metrics = null;
         switch (this._encoding) {
             case undefined:
             case '':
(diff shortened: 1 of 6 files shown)


source/view.js
```diff
@@ -3162,7 +3165,16 @@ view.TensorSidebar = class extends view.ObjectSidebar {
         if (value.type) {
             const item = new view.ValueView(this._view, value, '');
             this.add('type', item);
-            item.toggle();
+        }
+
+        if (value.initializer && value.initializer.category === 'Initializer') {
```
lutzroeder, line 3170: Is the category check used to gate against empty tensors?

source/view.js
```diff
@@ -4119,12 +4133,38 @@ view.Tensor = class {
     }
 
     get metrics() {
-        const metrics = Array.from(this._tensor.metrics || []);
-        const keys = new Set(metrics.map((metrics) => metrics.name));
-        if (!keys.has('sparisity')) {
-            // metrics.push(new view.Argument('sparisity', 0, 'float32'));
+        if (this._metrics === null) {
+            const value = this.value;
```
lutzroeder, line 4137: This leads to potentially large tensor values getting expanded immediately when the sidebar is opened. Would the metrics group need to be hidden initially via an expander or other UI?

source/view.js
```diff
@@ -4119,12 +4133,38 @@ view.Tensor = class {
...
+                    } else {
+                        num_zeros += Number(val === 0);
+                        num_parameters += 1;
+                    }
+                }
+
+                if (num_parameters > 0) {
+                    metrics.push(new view.Argument('sparsity', num_zeros / num_parameters, 'float32', 'percentage'));
+                } else {
+                    metrics.push(new view.Argument('sparsity', 0, 'float32', 'percentage'));
```
lutzroeder, line 4160: Is percentage more of a special case or are there other such metric styles that might be needed?

kylesayrs: Percentage would be used by other metrics such as operation sparsity and four block sparsity

source/view.js
kylesayrs: It's used to filter out some of the tensors for which metrics are not relevant/ not applicable. For example, `onnx::Unsqueeze_840` specifies the dimension of unsqueezing, metrics are not applicable here.

source/view.js
kylesayrs: I assume that when you refer to "sidebar" you're referring to "tensor properties sidebar". However,  I've disabled the automatic expansion of tensor values in the tensor properties sidebar [here](https://github.com/lutzroeder/netron/pull/1293/files/5fbe0128f05b7f5638f46e3e5b657f326987c5c4#diff-93b7b9ad1ecc0b6d6d5ef6461d90bad02ca0b903792489204364e1d656cbf8eaL3165).

In terms of having the metrics section in the tensor properties sidebar being hidden by an expander, I think that could make sense in the future when more metrics are implemented.

source/view.js
kylesayrs: As written, empty initializer tensors are still considered worthy of metrics, although maybe they shouldn't be.

lutzroeder: Got it. `Initializer` is not used consistently, across formats and operators these types of tensor cannot be determined generically. Another approach could be filtering vectors with small length. Is that worth special casing or up to the user to interpret the result.

source/view.js
lutzroeder: Not in the future, it will likely need to get addressed first. Accessing `this.value` can be an expensive operation and potentially lead to memory issues. For example, 500 MB tensor values in large models. 

Fundamental question is whether the Tensor sidebar should always expand content as the main value is showing content, render tensor visualization, compute metrics. In this is case the UX first needs to get hardened to indicate progress to the user and guard against errors. If not, all code paths accessing tensor content should be delayed by expanders.

Generally, the concern is less enabling metrics, more that it becomes a forcing function to prioritize the work to design and implement the Tensor sidebar, setting expectations with users that might need to get rolled back. There was no Tensor sidebar until now and the implementation is far from complete. For example, in many formats there is no away to activate the Tensor sidebar as there are no tensor names. Some formats have two layers making quantization part of tensor implementation. If usage increases these issues will need to get addressed or the feature disabled.

source/view.js
kylesayrs: > Initializer is not used consistently, across formats and operators these types of tensor cannot be determined generically

That makes sense.

A lot of the logic around "should metrics be applied to this tensor/node" must be done on a per-format basis, since you need the context of the operation to determine whether metrics are relevant. For example, it's difficult to tell whether sparsity applies to a tensor like `onnx::Unsqueeze_840` without knowing the context of ONNX and the unsqueeze operator.

As I see it, we have two design philosophies: In the case that the format does not provide a format-specific overload, then we can either *show default metrics* or *do not show default metrics*.

If we want to show default (sometimes unapplicable) metrics, we should just show them for all non-empty tensors and the user will know that formats with default metrics are not always reliable.
```javascript
if (!tensor.empty) {
```

If we don't want to show default metrics, then there's no point in implementing metrics on the `view.js` level. We should only be relying on format-specific implementations to provide metrics.

source/view.js
lutzroeder: Is percentage a special case of a `float32` value. `Argument` is a data concept and implemented pretty consistently. Introducing `style` would mix presentation with data concepts and this would potentially have to extend across all API layers. The alternative is handling it as a special `type`.

source/view.js
lutzroeder: In TensorFlow Lite for example, `Tensor` will contain the tensor data but also information to de-quantize. In other formats like Arm NN this happens in the `Value` wrapping an initializer. Tensor sidebar might need to evolve a set of features to support both a quantized and de-quantized view of the tensor data. The challenge is whether this can be generalized across the many different tensor formats and quantization approaches.

source/view.js
kylesayrs: > Argument is a data concept and implemented pretty consistently

You're correct, implementing the member `style` was a mistake. 

I initially avoided adding a special case of `type` since it's another case which has to be handled by functions which interpret `Argument`, for example #1241. However, on second thought, the fact that an `Argument` is a percentage also affects how it's interpreted.

In any case, I agree, this is probably best implemented as a special `float32-perc` style.

PR #1303, opened 2024-06-17
Colorize edge with int data type

seanshpark:
This will colorize edge with int data type.
This is to better visualize quantized models.

### DIFF
--- a/source/grapher.css
+++ b/source/grapher.css
@@ -69,6 +69,10 @@
 
 .edge-path { stroke: #000; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
 .edge-path-hit-test { pointer-events: stroke; stroke-width: 0.5em; fill: none; stroke: #000; stroke-opacity: 0.001; }
+.edge-path-b4 { stroke: #70e; stroke-width: 1x; fill: none; marker-end: url("#arrowhead"); }
+.edge-path-b8 { stroke: #04e; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
+.edge-path-b16 { stroke: #080; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
+.edge-path-b32 { stroke: #992; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
 
 .select > .node.node-border { stroke: #e00; stroke-width: 2px; }
 .select.edge-path { stroke: #e00; stroke-width: 1px; marker-end: url("#arrowhead-select"); }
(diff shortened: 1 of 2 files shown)


seanshpark:
sample capture of [Quantize_001.zip](https://github.com/user-attachments/files/15878340/Quantize_001.zip) tflite model


dark | light
--- | ---
image | image

source/grapher.css
```diff
@@ -69,6 +69,10 @@
 
 .edge-path { stroke: #000; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
 .edge-path-hit-test { pointer-events: stroke; stroke-width: 0.5em; fill: none; stroke: #000; stroke-opacity: 0.001; }
+.edge-path-b4 { stroke: #70e; stroke-width: 1x; fill: none; marker-end: url("#arrowhead"); }
+.edge-path-b8 { stroke: #04e; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
+.edge-path-b16 { stroke: #080; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
+.edge-path-b32 { stroke: #992; stroke-width: 1px; fill: none; marker-end: url("#arrowhead"); }
```
seanshpark, line 75: I'm not good at colors. Please choose with your preference :)

source/grapher.js
```diff
@@ -737,11 +737,53 @@ grapher.Edge = class {
         const createElement = (name) => {
             return document.createElementNS('http://www.w3.org/2000/svg', name);
         };
+        const getDatatype = (from) => {
+            if (from && from.value) {
+                const from_value = from.value;
+                let from_value_src;
+                if (from_value.outputs && from_value.outputs[0].value) {
```
kylesayrs, line 744:
```suggestion
                if (from_value.outputs && from_value.outputs[0]?.value) {
```

source/grapher.js
```diff
@@ -737,11 +737,53 @@ grapher.Edge = class {
...
+                            return value._type._dataType;
+                        }
+                    }
+                }
+            }
+            return '';
+        };
+        const edgePathClass = (dataType) => {
+            const dt = dataType.toLowerCase();
+            if (dt === 'uint8' || dt === 'int8') {
```
kylesayrs, line 765: nit: may consider swapping to a switch statement here

source/grapher.js
```diff
@@ -737,11 +737,53 @@ grapher.Edge = class {
         const createElement = (name) => {
             return document.createElementNS('http://www.w3.org/2000/svg', name);
         };
+        const getDatatype = (from) => {
+            if (from && from.value) {
```
kylesayrs, line 741:
```javascript
let from_value_src = from?.value?.outputs?.[0]?.value;
```

source/grapher.js
lutzroeder: Currently not use optional chaining. Check browser support matrix and apply consistently across the codebase.

lutzroeder (changes requested):
A few questions:

1. There are many more data types, why hardcode this set and how would this extend to other types?
2. Could this be overwhelming or confusing to other users?
3. What other solutions might work to solve this?

lutzroeder:
Closing as the pull request been stale for two weeks. Please comment if there are updates and it should be reopened.

PR #1311, opened 2024-07-09
Show "const input" layer as regular layer.

mkhsanov:
Fixed "const_input" layer type representation. Now it's not an input layer type but regular layer with all visible attributes.

image


lutzroeder:
@mkhsanov please share or add a sample file to reproduce this issue.

mkhsanov:
[llama.hn.zip](https://github.com/user-attachments/files/16157653/llama.hn.zip)

This model contains const_input layers.

PR #1312, opened 2024-07-09
Add support for .har files extension.

mkhsanov:
After some previous changes the netron application stopped to open .har files because this file extension isn't present in the official supported extensions list. Now it is able to open .har files also.

### DIFF
--- a/source/base.js
+++ b/source/base.js
@@ -1246,7 +1246,7 @@ base.Metadata = class {
             'gguf',
             'hd5', 'hdf5', 'keras',
             'tfl', 'circle', 'lite',
-            'mlnet', 'mar', 'maxviz', 'meta', 'nn', 'ngf', 'hn',
+            'mlnet', 'mar', 'maxviz', 'meta', 'nn', 'ngf', 'hn', 'har',
             'param', 'params',
             'paddle', 'pdiparams', 'pdmodel', 'pdopt', 'pdparams', 'nb',
             'pkl', 'joblib', 'safetensors',


lutzroeder:
`.har` is used by https://en.wikipedia.org/wiki/HAR_(file_format). Users more often select such files leading to errors while Hailo usage is small. One approach would be choosing a different file extension.

eldadr:
Hi @lutzroeder, it's Eldad from Hailo. I understand the conflict with the other file format with the same extension used for HTTP session recording. However, this recent change (removing the har extension support) breaks the compatibility of Hailo's toolchain with Netron, which probably means we will have to recommend our customers to use an old version of Netron (or a fork) to keep the har file support. The recent choice of Hailo as [Raspbrrery Pi's official AI accelerator](https://www.raspberrypi.com/news/raspberry-pi-ai-kit-available-now-at-70/), together with the recent [general availability](https://community.hailo.ai/t/dataflow-compiler-dfc-availability/1476) of Hailo's Dataflow Compiler, is expected to increase Hailo's usage.

What about a solution such as identifying the file type according to its content/magic in order to support both types with the same extension?

lutzroeder:
Hi @eldadr, can you outline how you are planning to resolve the underlying conflict? Is there a plan to rename the extension and could this be suggested to users as a temporary workaround until a fix is in place? Also, can you explain which scenario with toolchain compatibility needs to be addressed?

eldadr:
@lutzroeder, thanks for the reply, answering inline:

_can you outline how you are planning to resolve the underlying conflict?_

Add some mechanism that opens the file, checks its content, and then decides to which plugin to send it. The check can be very naive, for example decide that every file with har extension and tar archive content is a Hailo har, or we can add a more accurate check like making sure that there is a HN file inside the archive. @mkhsanov can comment with a more detailed proposal if needed, and also implement it, if you agree it's a good solution.

_Is the goal to rename the file extension and could this be suggested to users as a temporary workaround until a fix is in place?_

It can be used as a temporary workaround, for example we can register another extension such as har2 in Netron, and add a comment in our documentation that the users should rename their files from *.har to *.har2 whenever they want to use Netron. However, I think it will not be convenient in the long term and impact the user experience.

_Also, can you explain which scenario with toolchain compatibility needs to be addressed?_

The Hailo toolchain, also known as Hailo Dataflow Compiler, is a software package that converts ONNX/TFLite models to a binary file that runs on the Hailo hardware. This toolchain saves its intermediate results in the HAR format. So Netron is very useful in order to visualize these intermediate results, for example to compare the neural network before and after some optimization. If Netron doesn't support the HAR format, the users of the Hailo Dataflow Compiler will not be able to use it for this purpose, or it least it will be more difficult for them.

lutzroeder:
> users should rename their files from *.har to

Yes, moving to a file extension which isn't broadly used already would be a good long term solution. By analogy, if the format was called `.mp3` and over time Hailo `.mp3` model files gain in popularity, at which point is it appropriate to make a permanent change.

> doesn't support the HAR format

The app supports `.har` files via drag & drop and has the detection you describe in place. The question is more around making this easy to discover and the cost it incurs on others.

eldadr:
I understand, thanks for the feedback. We will consider it.

PR #1313, opened 2024-07-09
Remove duplicated inputs in case if input node connected with two or more nodes.

mkhsanov:
Remove duplicated inputs in case if input node connected with two or more nodes.

### DIFF
--- a/source/hailo.js
+++ b/source/hailo.js
@@ -108,7 +108,6 @@ hailo.Graph = class {
                 }
             }
         }
-
     }
 };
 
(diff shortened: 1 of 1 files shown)


lutzroeder:
@mkhsanov please add a sample file to illustrate the issue.

mkhsanov:
[VD_M_sRGB_4k_Apr03.har.zip](https://github.com/user-attachments/files/16157681/VD_M_sRGB_4k_Apr03.har.zip)

This archive contains .har model with input layer connected to two another layers. .har files extension support was fixed in one of previous pull requests - https://github.com/lutzroeder/netron/pull/1312

PR #1322, opened 2024-08-11
Use event delegation to reduce number of event listeners

beru:


source/grapher.js
```diff
@@ -811,24 +831,7 @@ grapher.Edge = class {
...
-                e.stopPropagation();
-            });
-        }
-        if (this.activate) {
-            this.hitTest.addEventListener('click', (e) => {
-                this.activate();
-                e.stopPropagation();
-            });
-        }
+        this.hitTest._var = this;
```
lutzroeder, line 834: Can this be accomplished without storing objects on HTML elements?

lutzroeder (changes requested):
@beru can you provide some context how to reproduce this issue. Can you test the change in different browser and desktop environments.

source/grapher.js
beru: I've updated the code. 2090f75554b5409170703ef28d5fe06016a1e59a

lutzroeder:
@beru can you rebase to main and resolve the build issues?

Issue #1308, opened 2024-06-30
Adding new model formats


lutzroeder:
This issue links discussions on adding new model formats.

Given the redundancy of many deep learning model formats, consider whether a new format is necessary. Explore adapting or improving existing formats to meet your needs.

PR #1333, opened 2024-08-29
Add `KaNN` format support

zylfried:
Hi 👋,

We've added support for `.kann` files to visualize KaNN graphs, a format developed by Kalray Inc. for accelerating ML computations with our MPPA processors.

Compatibility with Netron is essential for both our employees and customers but adapting an existing format do not address the specific requirements for handling KaNN's images and layers, among other things.

`kann.js`, `kann-metadata.json`, `kann-schema.js` have been added in `source/` as well as `kann.fbs` and the script to process it in `tools/`. Examples are also available in `test/models.json`.

Could you please review our pull request?

Let us know if you have any comments or need further information.

Thanks!


source/kann.js
```diff
@@ -0,0 +1,322 @@
+import * as flatbuffers from './flatbuffers.js';
+
+const kann = {};
+
+kann.ModelFactory = class {
+
+    // Check whether a KaNN file is processed and initialize the container
+    match(context) {
+        // Context is an object providing necessary information to interact with the processed file
+        const container = kann.Container.open(context); // attempt to get a container from the context
```
lutzroeder, line 10: This can mostly be removed and simplified as FlatBuffers detection in archives is mostly automatic. See `tflite.js` for an example. Are these files always `.tar.gz` without a file extension or mostly regular FlatBuffers files like TensorFlow Lite?

test/models.json
```diff
@@ -2157,6 +2157,38 @@
...
+    "type":     "kann",
+    "target":   "ResNet50-v1.5-MLPerf",
+    "source":   "https://github.com/kalray/netron-test-models/raw/main/ResNet50v15_MLPerf_quantized-fp16-5c.kann",
+    "format":   "KaNN",
+    "tags":     "quantization"
+  },
+  {
+    "type":     "kann",
+    "target":   "SqueezeNet-v1.0-ILSVRC2012",
+    "source":   "https://github.com/kalray/netron-test-models/raw/main/SqueezeNetV1_0-ILSVRC2012-fp16-5c.kann",
```
lutzroeder, line 2183: Can these be moved to an official model zoo repo that is being maintained.

test/models.json
```diff
@@ -2157,6 +2157,38 @@
...
+    "format":   "KaNN"
+  },
+  {
+    "type":     "kann",
+    "target":   "YOLO-v3-COCO",
+    "source":   "https://github.com/kalray/netron-test-models/raw/main/YOLO_v3-COCO_fp16_kv3_2_L2ON_onnx-fp16-5c.kann",
```
lutzroeder, line 2189: The local `target` names should have the same file extension as the server file. The linked files are in `.kann.tar.gz` format. Can you clarify if `.kann` files can be both a FlatBuffers file or a `.tar.gz` file or is this an optimization for these test files?

lutzroeder (changes requested):
> but adapting an existing format do not address the specific requirements for handling KaNN's images and layers, among other things.

Can you elaborate on what these requirements are? For example, would it be possible to expose these as operator extensions or metadata in ONNX instead of a having yet another mostly redundant ML graph format? What process is used to add this information to KaNN files?

package.js
```diff
@@ -548,7 +548,8 @@ const update = async () => {
         'sentencepiece', 'sklearn',
         'tf',
         'uff',
-        'xmodel'
+        'xmodel',
+        'kann'
```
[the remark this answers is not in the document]
zylfried: done

publish/electron-builder.json
```diff
@@ -62,7 +62,8 @@
         { "ext": "tnnproto",    "name": "TNN Model"                },
         { "ext": "torchscript", "name": "PyTorch Model"            },
         { "ext": "uff",         "name": "UFF Model"                },
-        { "ext": "xmodel",      "name": "Vitis AI Model"           }
+        { "ext": "xmodel",      "name": "Vitis AI Model"           },
+        { "ext": "kann",        "name": "KaNN Model"               }
```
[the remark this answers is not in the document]
zylfried: done

source/base.js
```diff
@@ -1254,7 +1254,7 @@ base.Metadata = class {
             'pkl', 'pickle', 'joblib', 'safetensors',
             'ptl', 't7',
             'dlc', 'uff', 'armnn',
-            'mnn', 'ms', 'ncnn', 'om', 'tm', 'mge', 'tmfile', 'tnnproto', 'xmodel', 'kmodel', 'rknn',
+            'mnn', 'ms', 'ncnn', 'om', 'tm', 'mge', 'tmfile', 'tnnproto', 'xmodel', 'kmodel', 'rknn', 'kann',
```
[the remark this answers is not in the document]
zylfried: done

source/kann.js
zylfried: They are always `.tar.gz` containing an inner FlatBuffers `.kann`

tools/kann.fbs
```diff
@@ -0,0 +1,60 @@
+namespace kann;
```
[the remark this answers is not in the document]
zylfried: done

test/models.json
zylfried: done

zylfried:
> > but adapting an existing format do not address the specific requirements for handling KaNN's images and layers, among other things.
> 
> Can you elaborate on what these requirements are? For example, would it be possible to expose these as operator extensions or metadata in ONNX instead of introducing yet another graph format? What process is used to add this information to KaNN files?

First, the current ONNX implementation doesn't meet the modularity needed to fully address KaNN's data type requirements. Additionally, not all KaNN layers have corresponding ONNX equivalents. For instance, custom layers like `cluster_split` or specific attributes of KaNN’s convolution layers don't exist in ONNX.
But most importantly, KaNN treats tensors as equivalent to layers, which necessitates a shift in how they are categorized. In KaNN, layers and tensors need to exist at the same hierarchical level, with Netron’s typical tensors replaced by arcs representing subviews of KaNN’s tensors.

It's important to clarify that KaNN is an inference engine similar to TensorRT, so the graphs we handle are more low-level than ONNX models. We have our own internal model format, and we'd like to use Netron for visualizing it but the difference in structure and purpose makes adapting existing formats insufficient to meet KaNN’s specific needs.

source/kann-metadata.json
```diff
@@ -0,0 +1,182 @@
...
+    {
+        "name": "WriteValue",
+        "category": "Data"
+    },
+    {
+        "name": "SiLU",
+        "category": "Activation"
+    },
+    {
+        "name": "ImageSMEM",
```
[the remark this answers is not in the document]
zylfried: As explained, KaNN handles tensors and layers differently than other implementations, so it would be valuable to keep model-specific operators to help users visually understand its structure.

test/models.json
zylfried: Our model zoo repo isn't public yet, but this repository is completely maintained and will be merged into the official zoo repo once it becomes public.

tools/kann.fbs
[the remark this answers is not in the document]
lutzroeder: Remove from `./tools`. Can the file be moved to an officially maintained repository for this format. e.g. "netron-test-models" is the official main location to learn about this format.

test/models.json
lutzroeder: Please front load this work to avoid churn and make it clear that these are official supported files not directly related to Netron.

source/kann-metadata.json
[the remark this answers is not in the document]
lutzroeder: Any operators which are not standard across formats. For example, "ImageSMEM" does not seem to be documented in online search. Its function or how it relates to other similar operators in other frameworks is not clear. Start by leaving it blank and create a follow-up issue in a KaNN repository to do such an investigation across formats and update categories for all similar operators in all formats.

source/kann.js
lutzroeder: Is there a guideline that such containers should always be named `.kann.tar.gz` or `.tar.gz`. Naming both the inner FlatBuffers file and the container `.kann` is confusing?

lutzroeder:
> the modularity needed to fully address KaNN's data type requirements

Can you provide an example?

> not all KaNN layers have corresponding ONNX equivalents

ONNX can act as a container with an entirely custom set of operators provided via a "domain".

> KaNN treats tensors as equivalent to layers

Can you elaborate in this. What is an "Arc" and is there some context why this is needed?

zylfried:
Indeed, we could extend ONNX to support KaNN’s data types and layer types, but the core difference lies elsewhere. KaNN’s purpose is not to introduce yet another framework for neural network design but rather to translate models from popular frameworks into a lower-level, processor-executable format. The `.kann` graph we aim to visualize in Netron represents an intermediate stage of model optimization and refactorization. This stage is then compiled into processor-level kernels, which is the final usable format but could not be visualized in Netron, the intermediate format being therefore used for visualization purposes only.

> > KaNN treats tensors as equivalent to layers
> 
> Can you elaborate in this. What is an "Arc" and is there some context why this is needed?

Unlike ONNX, where layers and tensors are distinct (layers as nodes and tensors as arcs), KaNN manipulates tensors as individual graph nodes, much like layers. The input/output relationships between these nodes are then not whole tensors but subviews (smaller parts of the tensors). These subviews correspond to the arcs in our graph, with an Arc being a connection between nodes, as defined in graph theory. Adopting the ONNX format would not work effectively and information would not be displayed properly (subviews' information being different from standard tensors').

> > the modularity needed to fully address KaNN's data type requirements
> 
> Can you provide an example or such a type or modularity need?

Although ONNX implementation does support many data types and could potentially meet our needs at one point, our KaNN implementation allows us to use any data type (whether as a single value or a list) at ANY point in the code without needing to specify individual compatibility for different parts (e.g. nodes, quantization...). This corresponds to our goal of improving our models in the near future to accept a wider variety of data types without having to modify Netron with each update.

Hope this helps you understand our need for a specific implementation of KaNN.

source/kann-metadata.json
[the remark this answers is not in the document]
zylfried: This would be `ImageSMEM` and `ImageDDR` then. However in relation with the global answer, those two operators are not actual layers but are necessary for Netron to visualize KaNN's unique tensor memory management at a lower level and will not be encountered in any other frameworks. They will be removed if it's non negociable, but doing so will be a significant loss of information for understanding our models.

source/kann.js
zylfried: The container isn't name `.kann` anymore, sorry for the confusion, but what should this guideline be ?

test/models.json
zylfried: In progress, but this will result in the models not being accessible for now

tools/kann.fbs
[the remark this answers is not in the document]
zylfried: In progress, but this will result in the `.fbs` not being accessible for now as the repo is currently private

source/kann-metadata.json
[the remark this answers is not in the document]
zylfried: Ok, this makes sense. Those operators are removed from metadata and we'll investigate how to categorize them properly for `ImageSMEM` and `ImageDDR` really need to be displayed distinctly.

source/kann.js
zylfried: done

source/kann-metadata.json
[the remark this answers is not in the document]
lutzroeder: "LRN" is category "Normalization".

zylfried:
Hi 👋,

The official model zoo repo is now public, and we should be in line with your requests for adding KaNN compatibility to Netron.

tools/kann
```diff
@@ -0,0 +1,36 @@
...
+sync() {
+    echo "kann sync"
+    mkdir -p "${third_party_dir}/metadata"
+    curl --silent --location --output "${third_party_dir}/metadata/kann.fbs" "https://github.com/kalray/kann-models-zoo/raw/refs/heads/main/utils/third-party/kann.fbs"
+}
+
+# regenerate kann-schema.js with 'tools/kann schema'
+schema() {
+    echo "kann schema"
+    [[ $(grep -U $'\x0D' ./source/kann-schema.js) ]] && crlf=1
```
lutzroeder, line 22: @zylfried running `./tools/kann sync schema` creates an invalid file, missing the namespace? Instead of a `third_party` folder in the model zoo can the `.fbs` file be move to a more official folder or repo with some format documentation.

lutzroeder (changes requested):

tools/kann
zylfried: Indeed the namespace was missing.
The thing about this `.fbs` file is that the format used for Netron isn't the final official serialized format as explained in previous answers. Since it's an intermediate serialization for now used for Netron visualization, keeping it in `third_party` seems appropriate. The official format is documented in our main repo.

lutzroeder: @zylfried can you elaborate and share a link to the previous answer you are referring to as well as a link to the documentation for the official format? Why is there a different serialization format instead of using the official one? Is this format only used for visualization purposes?

zylfried: Yes this is for now only used for visualization, here are some of my previous answers :

> It's important to clarify that KaNN is an inference engine similar to TensorRT, so the graphs we handle are more low-level than ONNX models. We have our own internal model format, and we'd like to use Netron for visualizing it but the difference in structure and purpose makes adapting existing formats insufficient to meet KaNN’s specific needs.

> KaNN’s purpose is not to introduce yet another framework for neural network design but rather to translate models from popular frameworks into a lower-level, processor-executable format. The `.kann` graph we aim to visualize in Netron represents an intermediate stage of model optimization and refactorization. This stage is then compiled into processor-level kernels, which is the final usable format but could not be visualized in Netron, the intermediate format being therefore used for visualization purposes only.

Full documentation is available on the model zoo repo if you're interested in our work [https://github.com/kalray/kann-models-zoo](https://github.com/kalray/kann-models-zoo)

lutzroeder: @zylfried instead of introducing another custom format only for visualization, what about exploring exporting to one of the existing 50+ formats the app supports or working on improving the built-in message format.

zylfried: I thought we'd already answered this topic, plus `.kann` or `.kgraph` are not temporary and may grow in importance over time but for now we'd like to use them for visualization to improve our process.

lutzroeder: If the user workflow requires exporting to a special format for visualization it is not clear this warrants maintaining yet another format. The deep learning format space is already very fragmented with often redundant formats that deviate only in implementation details. It wasn't clear that users have to create these files orthogonal to the framework in the context of visualization which makes it even less desirable as supporting the format wouldn't help streamlining the user workflow. Instead it might make more sense to export to a standardized format that supports all capabilities of Netron and can be shared by other frameworks. For example, exporting to ONNX as a container or  standardizing on visualization metadata seems a good starting point. `.message` was intended to represent an object model that reflects Netron capabilities and is used by some users for this purpose but isn't fully developed.

lutzroeder: The current implementation of arcs is not supported in Netron. e.g. using string manipulation to represent information that Netron can't process is not the right approach. The question seems more if there is a common denominator. For example, can the graph be presented in another format. The rendered graph that looks like an ONNX graph in the current implementation, there aren't any restrictions which operators ONNX can represent. Can extra information be treaded as metadata. For example, the current use of shape is not an actual shape, should it be annotated as metadata instead as it is not a construct the tool understands. Given the format is only used temporarily and there are many redundant formats already, what options have you explained to avoid introducing and maintaining a new format.

lutzroeder: > nodes are then not whole tensors but subviews (smaller parts of the tensors). These subviews correspond to the arcs in our graph, with an Arc being a connection between nodes, as defined in graph theory

Is there more documentation available explaining the motivations behind this design. The current implementation seems to reduce this to strings and is misusing tensor shapes, is the implementation actually relevant?

zylfried: We completely understand your concern about introducing yet another format to maintain, and we have explored adapting ONNX. However, it does not meet our specific requirements. Unlike ONNX and other formats supported by Netron, KaNN focuses on managing data across processor memory rather than operators. Without going into too much detail, images may be split into multiple tensors, stored in different memory regions, and processed in parallel by operators. To represent this in the graph, we need to manipulate strings not fully supported otherwise. Full documentation of the internal architecture isn't publicly available at this time.
Does this address your concerns? We are willing to address all your requests, but certain constraints in our format do not make it always possible.

lutzroeder: Makes sense. Generally it should be possible to export to a format like `message.js` or complete the implementation to include all features of the app. What remains are unsupported features. For example, `subview` is not a supported type in the app and multi-line text concatenation for tensor types is not supported. Such constructs would have to be documented and natively supported in the app or removed from the implementation.

zylfried: Thank you for your feedback and support. We understand your perspective, but what you’re suggesting now would require restarting work we thought had your initial agreement. We explored Netron’s constraints following your instructions (https://github.com/lutzroeder/netron/issues/1308 and more) and concluded a new format was necessary based on prior discussions. If `message.js` is now the preferred approach, earlier guidance would have been helpful to avoid months of back-and-forth. At this stage, starting over isn’t feasible for us. What do you think about integrating our current proposal?

lutzroeder:
Merged and removed changes which are incorrectly using string concat to work around features currently not supported.

zylfried:
Thanks a lot!
We'll investigate how to correctly bring to Netron what's missing.
Do you know when the next release will be available to all?

PR #1349, opened 2024-09-13
dagre.layout minor performance optimization

beru:
I checked speed difference using `npm run test measure` command. Compared to current main (3bf0ac7bf97b7446f2aa1b5c4a4eab4efcc62670), the `render:` line changed from 68.26 to 60.81. So I suppose there's minor performance improvement with this change.

<details><summary>npm run test measure log</summary>

```
# after
download:   0.07
load:      39.96
validate:  56.14
render:    60.81
total:    156.92

# before
download:   0.08
load:      42.89
validate:  60.29
render:    68.26
total:    171.43
```

</details>

### DIFF
--- a/source/dagre.js
+++ b/source/dagre.js
@@ -2234,9 +2234,9 @@ dagre.Graph = class {
         this.nodes = new Map();
         this.edges = new Map();
         if (this.compound) {
-            this._parent = {};
-            this._children = {};
-            this._children['\x00'] = {};
+            this._parent = new Map();
+            this._children = new Map();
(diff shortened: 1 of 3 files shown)


beru:
@lutzroeder 

Runnning `npm run test measure tvm/squeezenet.params` fails with the following output.

```
tvm/squeezenet.params
TypeError: blocks.values(...).map is not a function
```

This PR's commit 83ca32169fca9b9ad1547f485472ea6cc4f73e9d fixes the problem.
Without the change, `npm run test measure` doesn't finish.

Issue #1352, opened 2024-09-14
New automatic scroll to selected node is disorienting


fdwr:
- **Netron app and version**: 7.8.6 Desktop app
- **OS and browser version**: Win11 23H2 22631.4169, Edge 128.0.2739.67

## Steps to Reproduce:

1. Open a graph.
2. Click on a node to see its properties.
3. Notice the whole graph now shifts around, and the node you just clicked on is several nodes away now. This is pretty jarring and a little confusing, especially when you have parts of the graph that repeat and look similar to each other. I don't remember it doing this before?? 😵‍💫 Is there a hidden option to disable this or a way to use a previous version? 🥲

> Please attach or link model files to reproduce the issue.

Any.

fdwr:
It might be a collateral side effect of attempting to address this issue: https://github.com/lutzroeder/netron/issues/1344

beru:
@fdwr When a node is selected, the app scrolls the selected node to the center of the graph view ([link](https://github.com/lutzroeder/netron/blob/0d9daa1be341723f4bef53ffe22b0511b3b81cf6/source/view.js#L589-L607)) and I personally think it's [a little excessive](https://youtu.be/M2QQg3LonfI?t=11). By updating the code like this [link](https://github.com/beru/netron/commit/b41eea6e54f883e584946e06aafd123fb749b119), the scrolling occurs less often but it might be less exciting to scrolling addicts.

lutzroeder:
@beru can you create a pull request, ideally addressing the scrollbar issue.
@fdwr can you help test and review if the suggested change addresses the issue.

beru:
@lutzroeder `element.clientWidth` doesn't include width of scrollbar so I'm thinking to use this.
https://stackoverflow.com/a/40568748

Currently, the app scrolls a selected node to the center of the graph view. With the change I'm about to introduce with a new PR, the scrolling will not occur when a whole rectanngle area of selected node is in the container client area. I'm not really sure if this change of behaviour is acceptable but I'll open a PR anyways.

fdwr:
> @fdwr can you help test and review if the suggested change addresses the issue.

Sure, how? I just loaded ResNet into https://netron.app/, and the graph still shifts around on click 🤔. Might I need to explicitly refresh my browser cache?

beru:
[...]

PR #1354, opened 2024-09-14
Update view.View.scrollTo method

beru:
This PR is opened to address the issue reported in #1352

### DIFF
--- a/source/view.js
+++ b/source/view.js
@@ -588,6 +588,7 @@ view.View = class {
 
     scrollTo(selection, behavior) {
         if (selection && selection.length > 0) {
+            behavior = behavior || 'smooth';
             const container = this._element('graph');
             const bounds = container.getBoundingClientRect();
             let x = 0;
@@ -608,12 +609,11 @@ view.View = class {
                 selBottom = Math.max(selBottom, rect.bottom);
             }
             // If new selection is completely out of the bounds, scroll to centerize it.
-            if (selRight < bounds.left || selLeft > bounds.right || selBottom < bounds.top || selTop > bounds.bottom) {
+            if (selBottom - selTop > bounds.height || selRight - selLeft > bounds.width || selRight < bounds.left || selLeft > bounds.right || selBottom < bounds.top || selTop > bounds.bottom) {
                 x /= selection.length;
                 y /= selection.length;
                 const left = (container.scrollLeft + x - bounds.left) - (bounds.width / 2);
                 const top = (container.scrollTop + y - bounds.top) - (bounds.height / 2);
-                behavior = behavior || 'smooth';
                 container.scrollTo({ left, top, behavior });
                 return;
             }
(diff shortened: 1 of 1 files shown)


lutzroeder:
Would it help to have a buffer for nodes which are close to the visible border? For example, shrink the test rectangle by 10% so elements close to the border get scrolled into the visible area?

beru:
> 1. Would it help to have a border zone for elements which are close to corners of the visible area? For example, shrink the test rectangle by 10% so elements close to the corners also get scrolled into the visible area?

I'm not sure about that. Shall I create a commit to check UX?

> 2. Should elements scroll to center or only move into the visible area?

Centering a clicked node is a bit too much I think. I suppose moving it inside a [safe-area](https://en.wikipedia.org/wiki/Safe_area_%28television%29) will suffice. The size of the safe area can be around 80% or 90%.

If any of the attemps won't satisfy @fdwr, reverting c5868fe92cf2347749238a7bf8326da4bf270679 would be needed.

> 3. Long edges larger than the visible area cause scrolling when clicked. For example og_285 in [gpt2-lm-head-bs-12.onnx](https://github.com/onnx/models/blob/4c46cd00fbdb7cd30b6c1c17ab54f2e1f4f7b177/validated/text/machine_comprehension/gpt2-bs/model/gpt2-lm-head-bs-12.onnx).

Firstly, I'm not really sure if scrolling is necessary when edges are selected. Does it really help? In some cases when lines are very long, clicking them triggers a long scroll and connected nodes that were previously shown tend to go out of the visible area due to the scrolling.

beru:
@lutzroeder @fdwr done updating the code so can you please check new auto scroll behavior is OK?

lutzroeder:
1. Open [candy.onnx](https://raw.githubusercontent.com/Microsoft/Windows-Machine-Learning/master/Samples/FNSCandyStyleTransfer/UWP/cs/Assets/candy.onnx)
2. Open Find sidebar
3. Scroll to bottom and select `scale_B`
4. Scroll to bottom and select `scalePreprocessor`

**Actual:** does not scroll to new location.

beru:
@lutzroeder 

> 4. Scroll to bottom and select scalePreprocessor
> Actual: does not scroll to new location.

Fixed the bug with 19d547ffeaf29f1d127b1851f010b02bee9b0f05 by using `Number.NEGATIVE_INFINITY` instead of `0`. Currently, the change of this PR disregards your recent commit 7b993c63d1a04c806c45aed2eefbb0368b63e753 so a further adjustment might be needed...

This is off-topic but I've noticed that `outputImage` connection of [candy.onnx](https://raw.githubusercontent.com/Microsoft/Windows-Machine-Learning/master/Samples/FNSCandyStyleTransfer/UWP/cs/Assets/candy.onnx) isn't shown in the FIND sidebar list.

lutzroeder:
Regressed #1356

Open [alexnet.zip.pth](https://github.com/lutzroeder/netron/files/6096607/alexnet.zip.pth.zip).

Actual: initial positioning shows center of node
Expected: initial positioning should show top of node

beru:
> Regressed https://github.com/lutzroeder/netron/issues/1356

Honestly, it's difficult to change auto scroll behavior while keeping #1356.

PR #1357, opened 2024-09-16
Speed-up view.FindSidebar

beru:
This PR attemps to speed-up FIND sidebar by replacing SVG icons with font letters.

I used https://icomoon.io/app/ to create `netron.woff` font file. I also used https://iconly.io/tools/svg-convert-stroke-to-fill to convert strokes to fills in SVG files.

To check speed difference, [efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file can be used.

Rasterized shapes of icon letters are somewhat different from original SVG icons. So some adjustments would be needed...

<details><summary>Before/After</summary>

Before | After
-------|------
image | image

</details>


lutzroeder:
Can you provide some more context. How much is the improvement? Other options that do not require maintaining a font?

beru:
> Can you provide more context. How much is the improvement?

I recorded a screen to show the difference.

<details><summary>speed comparison</summary>

[efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file is used.

Before (037c36d796fecd2946233069e59981c7415837ad) | After (2ee4d8351284baf1d38316b363274db1fc4fbec7)
----|----
before | after

</details>

> Is maintaining a font the only option?

List Virtualization would be an effective solution for slow FIND sidebar. But implementing by myself takes time and efforts. 

Netron's current implementation uses so called [External symbol sprite](https://cloudfour.com/thinks/svg-icon-stress-test/#external-symbol-sprite), it seems like there're other faster methods but I guess using the icon fonts is the quickest.

> If there are adjustments needed, which are these and why are they not included in the pull request?

Compared to original SVG icons, Connections icon and Weights icon based on the icon font look thin. So I need to adjust its shapes using SVG editor software such as Inkscape. It's because of haste they aren't included in this PR.

beru:
I've updated the font. Now the icons look almost the same. Not pixel identical though.

<details><summary>Before/After</summary>

Before (7b993c63d1a04c806c45aed2eefbb0368b63e753) | After (2a34ef2ef789983833d96d1ffe212690e74c0d8c)
-------|------
image | image

</details>

beru:
Calling Document: createElement() method to create a lot of list items is time-consuming. So I've updated the code to use Node: cloneNode() method.

<details><summary>speed comparison</summary>

[efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file is used.

Before (037c36d796fecd2946233069e59981c7415837ad) | After (a97b4f32fc2a7a94e6e63e7031e360567e31d4f2)
----|----
before | after

</details>

I've also updated the font again. The icons look almost pixel identical now.

<details><summary>Before/After</summary>

Before (7b993c63d1a04c806c45aed2eefbb0368b63e753) | After (a97b4f32fc2a7a94e6e63e7031e360567e31d4f2)
-------|------
image | image

</details>

beru:
Here https://github.com/beru/netron/commit/56118a119fcbff1c668f8cae7d8b0880f3c70da6 is another version that doesn't use a web font. We can see that the other changes are still beneficial.

<details><summary>speed comparison</summary>

[efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file is used.

Before (037c36d796fecd2946233069e59981c7415837ad) | After (56118a119fcbff1c668f8cae7d8b0880f3c70da6)
----|----
before | after

</details>

beru:
While testing with [efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file, I've noticed that repeatedly opening and closing the FIND Sidebar makes the app screen broken. The problem reproduces with current `main` branch (7.8.8 Desktop App, Windows 10 22H2). This might be a bug or a limitation in Chromium. I don't see the same problem when the icons are drawn with the web font (this PR's code).

<details><summary>a screen recording</summary>

bug

</details>

Maybe I should open a new ticket to report this bug.

beru:
There's a different idea. Instead of creating a `svg` element per `li` element, creating a vertically long `svg` element that contains icons and overlaying it to the `ol` element would be more efficient.

lutzroeder:
> Here https://github.com/beru/netron/commit/56118a119fcbff1c668f8cae7d8b0880f3c70da6 is another version that doesn't use a web font. We can see that the other changes are still beneficial.

Let's switch to this approach if it has some of the same benefits (new pull request or replace this one). Custom font files are much harder to maintain and given this issue hasn't been reported widely seems an expensive trade-off. If there are other options we should try them as well.

The main improvement in this change is from cloning nodes? If so could this be accomplished by remembering the first node created which would avoid the need for `createDocumentFragment`?

beru:
> Let's switch to this approach if it has some of the same benefits (new pull request or replace this one).

I'll create a new pull request.

> Custom font files are much harder to maintain and given this issue hasn't been reported widely seems an expensive trade-off. If there are other options we should try them as well.

I understand so I'll try different approaches.

> The main improvement in this change is from cloning nodes?

I believe so. If I remember correctly, the use of `createDocumentFragment` didn't make much difference. I'll examine it.

> If so could this be accomplished by remembering the first node created which would avoid the need for createDocumentFragment?

I'm not really sure about that. My understanding is that when DOM nodes are added to on-screen DOM tree, it can cause frequent layout calculations. By using the [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment), that can be avoided/reduced.

> Creating and reusing existing SVG elements as items come into view or become invisible might be another idea?

I think that's what's called List Virtualization technique.

https://tghosh.hashnode.dev/rendering-large-lists-in-vanilla-js-list-virtualization

> Which part of using individual SVG icons makes this slow?

I've checked that use of `<svg>` icons in a lot of `<li>` list items is the main culprit.

<details><summary>42024 list item nodes</summary>

[efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file used.

image

</details>

> Is this happening on all machines and browsers or specific configurations?

This happenes on all Chromium-based browsers/applications.

Issue #1358, opened 2024-09-16
Repeatedly opening and closing Find sidebar makes app screen broken


beru:
- Netron app and version: 7.8.8, Desktop App and Chromium based browser
- OS and browser version: Windows 10 22H2, Ubuntu 24.04.1 LTS, Microsoft Edge 128.0.2739.79

Steps to Reproduce:

1. Open [efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file.
2. Open the FIND sidebar.
3. Close the sidebar
4. Repeat open/close until the problem occurs.

Please attach or link model files to reproduce the issue.

- [efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx)

lutzroeder:
@beru can you investigate why this is happening.

beru:
@lutzroeder 

> can you investigate why this is happening.

Chromium based browsers shows `Error code: Out of Memory` though my PC has 64.0 GB of installed RAM. This problem happens on both of Windows and Linux.

<details><summary>screenshots</summary>

## Google Chrome Version 128.0.6613.138 (Official Build) (64-bit)
![image](https://github.com/user-attachments/assets/19683283-538a-4c3c-aa7c-0f855a70e089)

## Microsoft Edge Version 128.0.2739.79 (Official build) (64-bit)
![image](https://github.com/user-attachments/assets/ee076aa5-fbbb-4bcb-ad52-75e7a9e4b13e)

</details>

lutzroeder:
@beru any insights why it is out of memory? Are there any structures not getting disposed?

beru:
> any insights why this is happening? Are there any structures not getting disposed?

I'm not sure since I still haven't debugged Chromium for this. I suspect this is a memory leak defect.

beru:
I've noticed that even without using a lot of `<svg>` elements in FIND sidebar, this problem eventually occurs after repeatedly open/close the sidebar.

<details><summary>screenshot</summary>

![image](https://github.com/user-attachments/assets/06aee036-157a-4ae1-b478-8981ad56a5f8)

<details>

PR #1359, opened 2024-09-19
Use Element: replaceChildren() method to remove all child nodes.

beru:
This PR replaces `innerHTML = ''` with `replaceChildren()` to remove all child nodes.

https://stackoverflow.com/a/65413839
https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren

`view.FindSidebar.deactivate` is also updated in the hope of resolving the problem reported in #1358. Although the change doesn't solve the problem, closing the sidebar gets faster.

<details><summary>recording</summary>

[efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file is used.

lifespan

</details>

### DIFF
--- a/source/view.js
+++ b/source/view.js
@@ -1513,7 +1513,7 @@ view.Menu = class {
     }
 
     _rebuild() {
-        this._element.innerHTML = '';
+        this._element.replaceChildren();
         const root = this._root[this._root.length - 1];
         for (const group of root.items) {
             const container = this._document.createElement('div');
@@ -2451,7 +2451,7 @@ view.Sidebar = class {
         sidebar.addEventListener('transitionend', (event) => {
             if (event.propertyName === 'opacity' && sidebar.style.opacity === '0') {
                 const content = this._element('sidebar-content');
-                content.innerHTML = '';
+                content.innerHTML.replaceChildren();
             }
         });
     }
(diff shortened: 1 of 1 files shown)


lutzroeder:
[replaceChildren](https://caniuse.com/?search=replaceChildren) is more recent and requires Chrome 86. Is the improvement from calling in `deactivate` and would iterating on `lastChild` with `removeChild` work?

beru:
> [replaceChildren](https://caniuse.com/?search=replaceChildren) is more recent and requires Chrome 86.

Added a polyfill function for that. https://github.com/lutzroeder/netron/pull/1359/commits/e8b72a52531030cca85d229430a389e62adbb4ef

> Is the improvement from calling in deactivate and would iterating on lastChild with removeChild work?

Yes, I've confirmed that doing so still works.

I've found out that since I've added `this._content.replaceChildren();` line in `view.FindSidebar.deactivate`, I need to add `view.FindSidebar.activate` method and call `this._update();` in there. And the code in `view.FindSidebar.focus` needed to moved to `view.FindSidebar.activate`. I hope this change is correct. Please examine it.

-----

I've noticed that when I double-click Connections/Weights list item on FIND sidebar or click a connection on the graph view, CONNECTION PROPERTIES/TENSOR PROPERTIES sidebar is opened. After closing the sidebar, FIND sidebar reappers. But this is not the case with Nodes list item. After closing NODE PROPERTIES sidebar, FIND sidebar doesn't reappear. I thought this is a bit incosistent but I'm not sure what is a proper behavior here.

PR #1372, opened 2024-09-30
Make graph panning smooth

beru:
This change makes graph drag pannning smooth on Chromium-based browsers.

[efficientdet-d3.onnx](https://github.com/phantrdat/onnx-efficientdet/blob/5efe7b1eb45d69f7ca4eef1022540aa8a2024ae8/efficientdet-d3.onnx) file can be used to check the change.

### DIFF
--- a/source/index.html
+++ b/source/index.html
@@ -22,7 +22,7 @@ button { font-family: -apple-system, BlinkMacSystemFont, "Segoe WPC", "Segoe UI"
 .center { position: absolute; margin: auto; top: 0; right: 0; bottom: 0; left: 0; user-select: none; -webkit-user-select: none; -moz-user-select: none; }
 .select { user-select: text; -webkit-user-select: text; -moz-user-select: text; }
 .graph { display: flex; height: 100%; width: 100%; overflow: auto; outline: none; touch-action: pan-x pan-y; }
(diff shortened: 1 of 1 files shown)


beru:
I've just created an issue ticket for slow panning of big inline SVG.
https://issues.chromium.org/issues/370730311

Using `filter: blur(0px);` style cause the problem reported in #1373 so the change has been erased from `main` branch.

Writing `this._element('canvas').style.filter = page ? '' : 'blur(0px)';` at [link](https://github.com/lutzroeder/netron/blob/eaf318367373c05cda273f6b764409a526503f6c/source/view.js#L230) can be a workaround. But the thing is this style hack is beyond my comprehension.

beru:
Setting `drop-shadow(0px 0px 0px #000);` has the same effect as `filter: blur(0px);`. And importantly, it doesn't cause the problem reported in #1373.

lutzroeder:
@beru can you add some context why this is improving load performance, test the change across browsers and platforms, and submit a pull request?

beru:
@lutzroeder 

> can you add some context why this is improving load performance

To do so, I need to debug Chromium and it would take much time. I've already reported the slow panning problem on Chromium issues but it seems like the rendering stutters doesn't reproduce on a googler's side. So there's a chance that I've been hallucinating and seeing things...

> test the change across browsers and platforms, and submit a new pull request?

I have several computers so I can do some testing but the question is would it be acceptable for you to use the CSS hack to avoid the Chromium issue?

PR #1385, opened 2024-10-24
URL parameter "/?file_path=" added to load local model files

0hDEADBEAF:
Hello!

I've been working with Netron for some time and there's a feature that I was willing to add: loading local model files, hosted on the server, using an URL parameter.
This is similar as the currently supported `/?url` parameter, allowing us to load a model file from the Internet. The `/?file_path=` is just as the same but tries to load it from the server local files instead.
By default, this feature is disabled (eg. no file will be accessed on the server) to keep the current Netron behavior. This can be enabled using a new parameter called `--allow-local-files` when starting the server.

At the moment, loading a local file on Netron requires a manual action (drag/drop or a click on the "Load model" button). Using an URL parameter instead allows scripts to automatically load them on Netron, this is the main benefit of my PR.

### DIFF
--- a/source/__init__.py
+++ b/source/__init__.py
@@ -24,6 +24,8 @@ def main():
     parser.add_argument('--verbosity',
         metavar='LEVEL', help='output verbosity (quiet, default, debug)',
         choices=[ 'quiet', 'default', 'debug', '0', '1', '2' ], default='default')
+    parser.add_argument('--allow-local-files',
+        help='Allows serving local files', action='store_true')
     parser.add_argument('--version', help="print version", action='store_true')
     args = parser.parse_args()
     if args.file and not os.path.exists(args.file):
@@ -33,7 +35,8 @@ def main():
         print(__version__)
         sys.exit(0)
     address = (args.host, args.port) if args.host else args.port if args.port else None
-    start(args.file, address=address, browse=args.browse, verbosity=args.verbosity)
+    start(args.file, address=address, browse=args.browse,
+       verbosity=args.verbosity, allow_local_files=args.allow_local_files)
     wait()
     sys.exit(0)
 
--- a/source/browser.js
+++ b/source/browser.js
@@ -172,6 +172,11 @@ host.BrowserHost = class {
                 }
             }
         }
+        const file_path = params.get("file_path");
+        if (file_path) {
+            this._openModelFromFile(file_path);
+            return;
+        }
         const gist = params.get('gist');
         if (gist) {
             this._openGist(gist);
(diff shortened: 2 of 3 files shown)


lutzroeder:
@0hDEADBEAF can you explain an example how this would be used.

0hDEADBEAF:
Well, if I'm not mistaken, using the existing `file` argument, you can only serve one local file at a time. If you want to serve another file, you'd have to either restart the server with that new file (meaning that the previous file wouldn't be able to be accessed anymore from the server) or start another server instance with that new file. My PR allows us to dynamically load any local model file from the server without restarting it. This is especially useful when several people need to work on the same server instance that has many private models.

lutzroeder:
Would using a generic web server work for this scenario? For example, what are example steps and commands you or your team would use and what are you trying to optimize for?

0hDEADBEAF:
I'm not sure to properly understand what you mean by "generic web server" here. However, I can give you few details about how the features I've added in this PR would be useful to us.

We generate a lot of models dynamically, models that we want to keep private (and therefore we cannot expose them to the Internet and access them through Netron with the `/?url` parameter). Those models are hosted on a shared server on which we have a Netron server running. Several people access the Netron interface through that server (http://the_server:8080).
We want everyone to be able to work on the numerous models that are automatically generated as the server is running. We don't want to restart the server each time a new model is generated as this is a process we deal with very often.
Using such feature is also pretty handy when it comes to share a model to someone: we just need to share the link (http://the_server:8080/?file_path=/models/auto_model178765.onnx) instead of passing the model file itself to everyone who needs it. Is is especially useful since the models are hosted on a server, which mean we would need to download them through (s)FTP first, then uploading on Teams to share it to others. It is much more convenient to share a link instead.
Moreover, if a model file gets updated, it just needs a F5 refresh on the browser instead of a server restart or a heavy process such as downloading the updated model through SFTP, sharing it to everyone else and making sure everyone is correctly using the updated version.

0hDEADBEAF:
This is a workaround that could probably work indeed. However, it feels a bit inconvenient to run two servers when we could just use one, such as my PR allows us to do so 🙁
I've fixed the remaining Pylint warnings 🙂

lutzroeder:
The trade-off is between "a bit inconvenient" for a few users capable of running a web server and the complexity and security concerns this creates for many other users. There might be other approaches worth considering, like running a single server and explicitly adding models to that instance via an API, especially if the server side evolves into doing more than just serving files. Currently evolving the app into a full web server and addressing related issues is not a priority. There are many well-built and maintained options available to accomplish this.

PR #1410, opened 2024-12-20
Fix PaddlePaddle IR empty `a_array` and idx is `0`

megemini:
In some special cases:

- 'a_array' could be an empty list
- 'idx' of tensor could be `0` (just like a placeholder)

This PR fixed these problems ~

Test model:
[empty_list.json](https://github.com/user-attachments/files/18210686/empty_list.json)

@lutzroeder Could you plz review this PR ~ Really thx!!! 😁


megemini:
cc @wanghuancoder @changeyoung98 @luotao1

PR #1416, opened 2025-01-17
Fix: Escape Special Characters in Metadata Replacement in server.py

gpradhan07:
**Description of Changes**

- Added a line to escape special characters in the meta variable using `meta = re.escape(meta)`before replacing it in the HTML content.

**Reason for Change**

- Prevents runtime errors caused by invalid escape sequences (e.g., \C) in the meta string during regex replacement.

**Impact**

- **Ensures** the application handles metadata safely and avoids crashing for certain model files.

### DIFF
--- a/source/python.js
+++ b/source/python.js
@@ -5460,7 +5460,15 @@ python.Execution = class {
                 const kind = lhs.kindOf(name);
                 switch (kind) {
                     case 'i':
-                    case 's': {
+                    case 'f':
+                    case 's':
+                    case 't': {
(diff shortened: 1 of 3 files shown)


gpradhan07:
Hi @lutzroeder Is that Fixed now ? #1416

lutzroeder (changes requested):
1. Instead of escaping `meta`, what about escaping `identifier` a a more local fix.
2. Can you please rebase to `main`.

Issue #1234, opened 2024-02-24
Extension to add tensor/operator metadata/annotation from external text file


mciprian13:
@lutzroeder Thank a lot for creating this tool. It is very useful for the day-to-day work and I am sure a lot of people are finding it useful as well. Kudos!
Two particular features I came across needing multiple times:

1) Display in the graph visualizer some custom annotation/metadata for tensors/operators that is not part of the model file/descriptor but comes from other sources such as tools or runtimes. Some examples:
- Performance statistics: Display for each operator, in addition to its attributes, some performance statistics such as latency, memory consumption, energy consumption, accuracy, etc that are provided from the model execution runtimes or other analysis tools.
- Accuracy statistics: Display for each operator the error for a custom impleementation of that operator relative to a reference.

2) Modify the styling of the tensors/operators such as color to encode visually the above information. For example:
- Color each operator with a heatmap color gradient from blue to red to encode information such as: high error (red), low error (blue) or high computation efficiency (blue), low computation efficiency (red), etc.

Since each model format has a way to identity uniquely tensors and operators, the above two features could be realized by importing into Netron a text file or something (separate from the model file) that capture the meta we want, could be something like:

`tensor_id: <tensor name or index>`
`tensor_meta: <text string that will be added into the tensor description in the visualizer>`
`tensor_style: <syntax to specify the styling of the tensors - the arrows between operators - such as tickness, color, etc>`
`....`
`operator_id: <operator name or index>`
`operator_meta: <text string that will be added into the operator text description in the visualizer>`
`operator_style: <syntax to specify the styling of the operators - the boxes - such as background color, font, etc>`
`....`

Would also be nice if the above source file can be provided via command line such that Netron will open automatically a model together with its meta file.

I understand different formats have different ways to identity tensors/operators. For example I mainly use TensorFlowLite in my daily work where tensors/operators are uniquely identified with an index (called "Location") but there's also ONNX where tensors/operators are identified via strings/names (I don't know whether this guarantees uniqueness). But maybe this mechanism is not that hard to be generalized.

In summary the above features allows different users to add other custom information into the visualizer.
Currently we had to do our own visualizer (using graphviz) just to have the two features above.
Would be very useful for the Neutron community to have these features embedded into the tool.

LATER EDIT: I was thinking about how the meta text could be inserted and I see two ways:
- Either add in the box itself
[...]

PR #1418, opened 2025-01-17
Extension to visualize metadata provided externally

leman31:


lutzroeder (changes requested):
@leman31 can you share some model and metadata files to test this implementation. How are the metadata files generated? Can you also start reworking the code from directly modifying the DOM to align with the underlying object model. Security related removals need to be reverted. Assuming you read the threads attached to #1240 and #1241?

mciprian13:
@lutzroeder Hello, I am the tech lead of a team in NXP and I decided to ask a student (@leman31) to implement this feature which helps us a lot in model debugging purposes.
A reminder: we discussed extensively about this feature [here](https://github.com/lutzroeder/netron/issues/1234).
I will not go again through all the details, just a summary:
- This feature allows to add extra annotations on top of a model that is provided through a separate file (not the model itself).
- This allows to add annotations on both tensors and operators, such as:
  - Metadata as key-value pairs.
  - Actions (button) that trigger specific shell commands or javascripts.
  - Opening a **text** or **image** when hovering over the tensor/operator with the mouse.
 With these very simple, generic but versatile commands a lot of functionalities can be obtained.

To be noted that this mechanism was tested only with TFLite models (which was our usecase). This has to do with how we reference tensor/operator IDs in the metadata file using the integer **identifier** field which is TFLite specific. Maybe this can be generalized easily to other formats as well.

Please find in the attached archive what you are looking for:
[Package.zip](https://github.com/user-attachments/files/18619053/Package.zip)
You can find:
- A readme which describes how the feature works
- Some tests including a TFLite model and some metadata text files (some of them might contain some invalid stuff for negative testing or some invalid paths ... but you should get a quick feel how it all works and change it). These are simple examples to understand how it works.
Maybe you can advise us where is the best place to put that document (Readme).

The metadata text file in our specific case is generated by **our** testing infrastructure (which is custom/internal). In our specific case it is extremly useful to export information from the runtime related to the errors we get at each tensor (tensor metrics as text and tensor error heatmaps as images). Thus we annotate the graph and we can see extra text and images for each tensor to know whether **our** runtime produces extra errors relative to vanilla TFLite runtime. This is super cool and very useful and I am really confident that others will consider this useful for **their** own purposes.

Again, I would not insist on how these files are produces and trying to impose a standard because, in principle, the metadata text file syntax is very simple and the features are very generic and thus anyone who likes these features and find them useful can use them easily in their own infrastructure. We are open to changing the syntax of the metadata file if you want although we chose the simplest syntax possible.

We would love to contribute this feature to the community and even improve it based on feedback with the only request that the functionality is maintained (afterall this is the philosophy of open-source).

Once we decide the strategy, we can discuss the technical parts (go on with the code review, update based on comments, etc).
What do you think?

lutzroeder (changes requested):
@mciprian13 @leman31 thank you for sharing the sample files. Let's start by getting to a point where I can try out the feature. 1. Fix the build errors 2. starting the app currently does not show the feature (on macOS) and 3. opening via browser is showing a black screen.

screenshot

The pull request also removed multiple security checks and these changes need to get reverted.

As mentioned earlier, modifying the DOM tree directly and side stepping the object model might need to be addressed differently, let's discuss once the pull request is in a working state.

mciprian13:
@lutzroeder Indeed, we tested only the application executable and also the web application on Windows. We will try to get familiar with the CI/CD jobs for the other environments.
We can send you a windows exe installer if you want to try out the features (or a video) :)
We removed some of the code because the features were not working in the browser. I totally agree, we need to revisit that.
Thanks for your availability. For now I need to wait for the student to return from the exam session to continue.
Will keep you posted.

mciprian13:
@lutzroeder The build problems were fixed.
Please have a look on the features by looking into the provided archive (one of the links above):
https://github.com/user-attachments/files/18619053/Package.zip
We can continue the discussions on the features, code review, etc.
Thank you!

mciprian13:
@lutzroeder Can you please have another look? We are available to answer your questions and requests related to this PR. Thank you!

mciprian13:
@lutzroeder Kind reminder to help us, based on your availability, to review and land this feature which is important to us.

lutzroeder:
Hi @mciprian13 @leman31 — to help move this forward:

* Let's revert all security-related changes, including the use of `unsafe-inline` and `unsafe-eval`.

* The idea of enabling actions is interesting. However, allowing arbitrary executables to be invoked based on a metrics file introduces significant security concerns — e.g., could a metrics file be crafted to execute malicious code? Also, referencing specific executables isn't portable across platforms. Have you considered using protocol handlers instead?

* The current pull request modifies and cleans HTML directly, rather than building on the existing data model and sidebar integration. I pushed a change to main that extends the current metrics implementation to support external `.metrics.json` files using the existing sidebar logic. Could we shift to that approach as a baseline?

For example, if you open the attached `.onnx` file in main, then drag a `.metrics.json` file into the view, its metrics will appear in the sidebar alongside built-in and model-derived metrics.

[mnist.metrics.zip](https://github.com/user-attachments/files/19698325/mnist.metrics.zip)

mciprian13:
> Hi @mciprian13 @leman31 — to help move this forward:
> 
> * Let's revert all security-related changes, including the use of `unsafe-inline` and `unsafe-eval`.
> * The idea of enabling actions is interesting. However, allowing arbitrary executables to be invoked based on a metrics file introduces significant security concerns — e.g., could a metrics file be crafted to execute malicious code? Also, referencing specific executables isn't portable across platforms. Have you considered using protocol handlers instead?
> * The current pull request modifies and cleans HTML directly, rather than building on the existing data model and sidebar integration. I pushed a change to main that extends the current metrics implementation to support external `.metrics.json` files using the existing sidebar logic. Would it be possible to shift to that approach as a baseline?
> 
> For example, if you open the attached model file and drag a `.metrics.json` file into the view, its metrics will appear in the sidebar alongside built-in and model-derived metrics.
> 
> [mnist.metrics.zip](https://github.com/user-attachments/files/19715661/mnist.metrics.zip) [model.metrics.zip](https://github.com/user-attachments/files/19715662/model.metrics.zip)

To your points:
1) We can revert those changes although we will have to reevaluate to see whether some features are impacted or not.
2) Having actions by running executables/commands is a very powerful way to add all sort of functionality. For us the security aspect is irrelevant because we are creating the meta files and we are consuming the meta files. The meta files are ways to add extra information/functionality to Netron from outside. This is good enough for us.
3) Having the same meta in JSON format is a small change but changing the way the same information is imported in the tool is a very significant change and I am afraid we don't have time for such a big change. We will have to evaluate the effort needed but if it is too large we will have to drop the path of upstreaming this.

PR #1419, opened 2025-01-19
update DaVinci OM SVP version

wujunliang00:
1. Add version check to determine whether it is a release or debug. 
2. SVP = Smart Vision Platform

### DIFF
--- a/source/om.js
+++ b/source/om.js
@@ -466,7 +466,7 @@ om.Container = class {
                     break;
                 }
                 case 'PICO': {
-                    this.format = 'DaVinci OM SVP'; // SVP = Smart Vision PICO
+                    this.format = 'DaVinci OM SVP'; // SVP = Smart Vision Platform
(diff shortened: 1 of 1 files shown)


lutzroeder:
@wujunliang00 where is the `version` property used?

wujunliang00:
> @wujunliang00 where is the `version` property used?

this version property is used to determine whether it is a debug or release version. If it is a debug version, it will include the model graph, while the release version does not contain the model graph. The release version is intended for direct deployment, so the model graph is reduced to minimize runtime.

lutzroeder:
@wujunliang00 where in the code is `this.version` used and shown to the user? Can you please share a sample file and screenshot.

wujunliang00:
> @wujunliang00 where in the code is `this.version` used and shown to the user? Can you please share a sample file and screenshot.

截屏2025-01-30 19 09 57
Model properties will be shown to the user

PR #1422, opened 2025-01-23
show json in attributes

xieofxie:
I think this simple approach is enough to use?

- Add an open external data menu
- Add a new attribute in node like ext-data
- Add a Show / Hide External Data

So we could show external data in the node

The input json will be in the format
{
  "node name1": {  "ext1": 123, "ext2": 456 },
  ...
}

### DIFF
--- a/source/browser.js
+++ b/source/browser.js
@@ -441,12 +441,34 @@ host.BrowserHost = class {
         return await this._openContext(context);
     }
 
+    async _readFileAsText(file) {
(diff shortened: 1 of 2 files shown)


xieofxie:
related issue https://github.com/lutzroeder/netron/issues/1241

lutzroeder:
@xieofxie see discussion in #1240 and #1418.

PR #1425, opened 2025-01-29
KaNN: add missing data

zylfried:
Hi,
We've quickly added some data that weren't correctly extracted in the previous code.
Feel free to review our changes.


lutzroeder (changes requested):
@zylfried can you add a small test file. If one of the existing test files is using quantization add an assert check. Is this using `linear` quantization instead of a generic `annotation`?

PR #1436, opened 2025-03-03
Add copy name button to side bar

acmore:
This PR is to add a copy action to the names of the node and arguments.


lutzroeder:
@acmore adding another button to most text fields adds quite a bit of clutter to the user interface while the same can be accomplished via select and copy. How could this be solved differently? Have you explored alternatives?

acmore:
> @acmore adding another button to most text fields adds quite a bit of clutter to the user interface while the same can be accomplished via select and copy. How could this be solved differently? Have you explored alternatives?

Yeah, select and copy can work by select the text carefully when the name is longer. But there are some issues:
1. When the name is longer, it's a bit difficult to select
2. The text selected in this way included a new line
3. Sometimes will click through the element when it's a tensor

I do agree that it's a bit redundant, I am good to keep it in my own branch :)

PR #1438, opened 2025-03-07
OpenVINO: unspecified and dynamic type support

EgorDuplensky:
After recent changes in OpenVINO it is possible to get
'unspecified' and 'dynamic' types when dumping IR

Synthetic IR examples:
- unspecified:  
[unspecified.gz](https://github.com/user-attachments/files/19128549/unspecified.gz)
- dynamic
[dynamic.gz](https://github.com/user-attachments/files/19128553/dynamic.gz)


EgorDuplensky:
@lutzroeder I have added synthetic IRs with those data types into models.json.
Let me know if I need to do it differently

PR #1467, opened 2025-06-13
add dockerfile

2niuhe:
Add dockerfile.


lutzroeder:
@2niuhe can you provide some context why this change is needed and how it is being used?

PR #1519, opened 2025-12-03
Support espdl 

sun-xiangyu:
[ESP-DL](https://github.com/espressif/esp-dl) is a deep learning library developed by Espressif Systems for deploying neural networks on resource-constrained microcontrollers. The model extension is `.espdl`.
Since Netron is an excellent tool for visualizing neural network models, we hope to contribute by integrating ESP-DL model support into it.

For this pull request, we have implemented all necessary visualization functions, for example:

- Load .espdl model files.
- Present all layer-related parameters in the sidebar.
- Visualize the entire model as a computation graph.

@lutzroeder , please help review and meger it.

### DIFF
--- a/source/espdl.js
+++ b/source/espdl.js
@@ -1,5 +1,4 @@
 import * as flatbuffers from './flatbuffers.js';
-import * as zip from './zip.js';
 
 const espdl = {};
 
@@ -37,8 +36,7 @@ espdl.ModelFactory = class {
             throw new espdl.Error(`Failed to load ESPDL schema: ${error.message}`);
         }
         let model = null;
-        const attachments = new Map();
-        if (context.type == 'espdl.binary') {
+        if (context.type === 'espdl.binary') {
             // Read from stream directly
             const stream = context.stream;
             if (!stream) {
@@ -104,7 +102,7 @@ espdl.Model = class {
 
 espdl.Graph = class {
 
-    constructor(metadata, graph, model, stream) {
+    constructor(metadata, graph) {
         this.name = graph.name || '';
         this.inputs = [];
         this.outputs = [];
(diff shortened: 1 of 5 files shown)


source/view.js
```diff
@@ -6860,7 +6861,6 @@ view.ModelFactoryService = class {
                 { name: 'CviModel data', value: /^CviModel/ }, // https://github.com/sophgo/tpu-mlir/blob/master/include/tpu_mlir/Builder/CV18xx/proto/cvimodel.fbs
                 { name: 'DRTcrypt data', value: /^DRTcrypt/ },
                 { name: 'ELF executable', value: /^\x7FELF/ },
-                { name: 'EDL2 data', value: /^EDL2/ },
```
sun-xiangyu: The ESPDL model starts with the string "EDL2", so I removed this line; I'm not sure if you had any additional considerations.

source/espdl.js
```diff
@@ -0,0 +1,602 @@
...
+                    // We'll set a custom context type, but we need to skip header later
+                    // For now, just indicate match
+                    return context.set('espdl.binary', null); // We'll read the buffer in open method
+                }
+            }
+        }
+        // Also try to parse as flatbuffers (maybe header already stripped)
+        const reader = await context.peek('flatbuffers.binary');
+        if (reader) {
+            // No file identifier expected for ESP-DL flatbuffers
```
lutzroeder: Add an identifier to the format?

source/espdl.js
```diff
@@ -0,0 +1,602 @@
...
+        return new espdl.Model(metadata, model, stream);
+    }
+};
+
+espdl.Model = class {
+
+    constructor(metadata, model, stream) {
+        this.format = 'ESP-DL';
+        if (model.ir_version !== undefined) {
+            this.format = `${this.format} IR v${model.ir_version}`;
```
lutzroeder: Is `ESP-DL` a different format from `ESP-DL IR` or should this be `ESP-DL v1`?

lutzroeder (changes requested):
> None of the existing formats (ONNX, TFLite, Caffe, …) can describe all of these tricks at the same time, so we would have had to carry an out-of-tree “annotation” file anyway.

Could this be accomplished by using one of the extensible metadata mechanisms formats offer. ONNX for example has `metadata_props` for most objects. Operators could still use their own conventions. This would avoid inventing yet another graph container format.

PR #1521, opened 2025-12-04
Add TOSA support

fresve-arm:
Added support for TOSA.

See: https://www.mlplatform.org/tosa/tosa_spec.html


source/base.js
```diff
@@ -1,3 +1,4 @@
+// SPDX-FileCopyrightText: Copyright 2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
lutzroeder: Remove

LICENSE
```diff
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) Lutz Roeder
+Copyright 2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
```
lutzroeder: Remove

package.js
```diff
@@ -611,7 +612,8 @@ const update = async () => {
         'sentencepiece', 'sklearn',
         'tf',
         'uff',
-        'xmodel'
+        'xmodel',
```
lutzroeder: Sorted alphabetically

tools/tosa_metadata.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright 2025 Arm Limited and/or its affiliates <open-source-office@arm.com>
+
+"""Parsing of tosa.xml to tosa_metadata.json"""
```
lutzroeder: Since this doesn't have any Python specific dependencies please converted to JavaScript using `../source/xml.js`.

tools/tosa
```diff
@@ -0,0 +1,62 @@
...
+        curl --silent --show-error --location "https://raw.githubusercontent.com/arm/tosa-specification/refs/heads/${TOSA_VERSION}/tosa.xml" -o ./third_party/tosa/tosa_${TOSA_VERSION}.xml
+    done
+}
+
+install() {
+    echo "tosa install"
+}
+
+schema() {
+    for TOSA_VERSION in ${TOSA_VERSIONS[@]}
```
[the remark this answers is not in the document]
fresve-arm: I'm willing to drop pre-1.0 support if this would ease the acceptance of this PR? 
With 1.0 and forwards there should not be any breaking changes to the schema, which would reduce the need for maintenance.

mciprian13:
@fresve-arm This is super nice!
We also have the same need (and the rest of the industry) for TOSA.
TOSA is starting to gain a lot of traction but unfortunatelly there is no good visualizer for it.
Currently only the ModelExplorer supports it but with an extra adapter which in my opinion looks ugly:
https://github.com/google-ai-edge/model-explorer
Supporting TOSA is a must these days!
I am looking forward to seeing this merged. Thanks!

PR #1531, opened 2026-01-02
Find sidebar node type filter

tnzl:
# Problem

When working with large ONNX models, the search sidebar can list thousands of nodes. Filtering by name works, but it's hard to focus on specific operation types (e.g., finding all Conv, MatMul, or GroupQueryAttention nodes). Users had to scroll through many results or type exact names, which is slow for exploratory work.

## Solution
image

Add a node type filter in the search sidebar for ONNX models. A filter icon button appears next to the search input, matching the existing toggle button style. Clicking it opens a dropdown with all unique node types found in the model. Selecting a type filters the results to show only nodes of that type, and it works together with text search for more precise filtering.

### Key Features:

- **Filter icon button**: A filter icon (three horizontal lines) appears in the search bar for ONNX models, styled like the existing toggle buttons

- **Dynamic node type collection**: Automatically collects and lists all unique node types from the model (e.g., Conv, MatMul, Add, GroupQueryAttention, Cast, Concat, etc.)

- **Dropdown menu**: Clicking the filter icon opens a dropdown with:
  - "All Types" option to clear the filter
  - Sorted list of all unique node types
  - Visual indication of the selected type (blue color, bold text)
  - Hover effects for better UX

- **Combined filtering**: Works with the existing text search — users can filter by type and search by name simultaneously

- **ONNX-only**: Only appears when viewing ONNX models (detected by checking if model format starts with "ONNX")

- **State persistence**: Selected filter state is preserved when the sidebar is reopened

### Technical Changes:

- Added `nodeType` to the search state object
- Implemented `_collectNodeTypes()` method to extract unique node types from the model graph
- Modified `_node()` filtering logic to respect the node type filter
- Added filter icon SVG symbol (`sidebar-icon-filter`) to match existing icon style
- Implemented dropdown menu with proper event handling and click-outside-to-close behavior
- Added CSS styling with dark mode support

@lutzroeder , long time user, first PR,  please help review.

### DIFF
--- a/source/view.js
+++ b/source/view.js
@@ -4319,30 +4319,94 @@ view.FindSidebar = class extends view.Control {
         this._query = this.createElement('input', 'sidebar-find-query');
         this._search.appendChild(this._query);
         
-        // Add node type filter dropdown for ONNX models
+        // Add node type filter button for ONNX models
         if (this._isOnnxModel) {
-            this._nodeTypeSelect = this.createElement('select', 'sidebar-find-nodetype-filter');
-            this._nodeTypeSelect.setAttribute('title', 'Filter by Node Type');
-            const defaultOption = this.createElement('option');
-            defaultOption.value = '';
+            const container = this.createElement('div');
+            container.style.position = 'relative';
+            
+            this._nodeTypeFilterButton = this.createElement('label', 'sidebar-find-toggle');
+            this._nodeTypeFilterButton.innerHTML = `<svg class='sidebar-find-toggle-icon'><use href="#sidebar-icon-filter"></use></svg>`;
+            this._nodeTypeFilterButton.setAttribute('title', 'Filter by Node Type');
+            
+            // Create dropdown menu (initially hidden)
+            this._nodeTypeDropdown = this.createElement('div', 'sidebar-find-nodetype-dropdown');
+            this._nodeTypeDropdown.style.display = 'none';
+            
+            const defaultOption = this.createElement('div');
+            defaultOption.style.padding = '6px 12px';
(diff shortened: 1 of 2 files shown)


Horlabrainmoore (approved):

source/view.js
```diff
@@ -4802,13 +4072,15 @@ view.FindSidebar = class extends view.Control {
...
-            weight: true
+            weight: true,
+            nodeType: ''
         };
         this._toggles = {
             node: { hide: 'Hide Nodes', show: 'Show Nodes' },
             connection: { hide: 'Hide Connections', show: 'Show Connections' },
             weight: { hide: 'Hide Weights', show: 'Show Weights' }
         };
+        this._isOnnxModel = this._view._model && this._view._model.format && this._view._model.format.startsWith('ONNX');
```
lutzroeder: Can you please implement this to work for all model formats.

lutzroeder (changes requested):

OPS-NeoRetro:
@tnxl your idea works really well! @lutzroeder was right about you only doing this for ONNX, because imagine how easier lives are when people use Netron for TFLite and TensorFlow Saved Model files. Since your branch is only named `feature/onnx-node-type-filter` but you didn't say that this is only for ONNX in the title, it's the reason @lutzroeder wanted you to support other formats. The dropdown feels a little like a bolted-on component though. This is my experience from a video that I made myself

https://github.com/user-attachments/assets/2404c7d6-0a4f-4977-93b6-df454db62f84

@lutzroeder should we also add a help page to netron.app?

PR #1538, opened 2026-01-30
To support click Automatically copy name

alan-chen-dongsheng:
when user select a node or edge, will auto copy the name

image

> why I want this
> The reason for this demand is that many AI compiler engineers, like me, are currently in the process of manual intervention optimization because node names in the network need to be duplicated frequently

### DIFF
--- a/source/desktop.mjs
+++ b/source/desktop.mjs
@@ -155,6 +155,9 @@ desktop.Host = class {
     }
 
     async start() {
+        electron.ipcRenderer.on('copy', () => {
+            this._view.copy();
+        });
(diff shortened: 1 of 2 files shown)


lutzroeder:
@sexjun can this be implemented as click and then <kbd>Cmd</kbd>+<kbd>C</kbd>? The current change is not a common UI pattern and likely confusing to many users.

sexjun:
@lutzroeder I think what you said makes sense, please check it

lutzroeder:
@sexjun can you help review and test the change in main.

PR #1562, opened 2026-04-20
Keyboard navigation

JoMazM:
Added different keyboard navigation shortcuts in the Node view:
- Arrow keys used to pan in the navigation view.
- WS keys to zoom in/out
- Ctrl + Arrow keys to navigate between nodes.

### DIFF
--- a/source/view.js
+++ b/source/view.js
@@ -2497,12 +2497,16 @@ view.Graph = class extends grapher.Graph {
             this._events.gesturestart = (e) => this._gestureStartHandler(e);
             this._events.pointerdown = (e) => this._pointerDownHandler(e);
             this._events.touchstart = (e) => this._touchStartHandler(e);
+            this._events.keydown = (e) => this._keydownHandler(e);
+            this._events.click = (e) => this._clickHandler(e);
             const document = this.host.document;
             const element = document.getElementById('target');
             element.focus();
             element.addEventListener('scroll', this._events.scroll);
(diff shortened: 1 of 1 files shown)


source/view.js
```diff
@@ -2697,6 +2716,131 @@ view.Graph = class extends grapher.Graph {
...
+                if (e.ctrlKey) {
+                    this._navigateNodes('right');
+                } else {
+                    container.scrollLeft += panAmount;
+                }
+                handled = true;
+                break;
+            case 'w':
+            case 'W':
+                this.zoom *= 1.1;
```
lutzroeder, line 2759: There are existing zoom shortcuts. Would it make sense to reuse those instead?

source/view.js
```diff
@@ -2697,6 +2716,131 @@ view.Graph = class extends grapher.Graph {
...
+        const document = this.host.document;
+        const container = document.getElementById('target');
+        const panAmount = 50;
+        let handled = false;
+        switch (e.key) {
+            case 'ArrowUp':
+                if (e.ctrlKey) {
+                    this._navigateNodes('up');
+                } else {
+                    container.scrollTop -= panAmount;
```
lutzroeder, line 2729: Arrow keys already scroll the container when focused. Is there a case where this is needed?

lutzroeder (changes requested):
Thank you. The node navigation is a great idea. Could it be implemented using the existing `activate`, `select`, and `scrollTo` methods?

PR #1568, opened 2026-04-22
Add session-scoped graph visibility controls and editable labels

spsrathor:
Add a local editing layer to Netron’s graph viewer so users can
interactively control what is shown after loading a model, without
changing the underlying model file.

This change introduces a session-scoped visibility system in the view
layer. Users can now hide or show whole nodes, node-level categories
such as names, attributes, inputs, outputs, weights, and connections,
and individual attributes from the node properties sidebar. Visibility
changes are applied to the rendered graph immediately by rebuilding the
view graph from UI state rather than mutating parser output.

The graph construction path in source/view.js now respects per-session
visibility state during node creation and edge creation. Hidden nodes
are skipped from the rendered graph, and connections are suppressed when
their source or destination visibility rules disable them. Existing
global options for names, attributes, and weights remain supported and
act as defaults, while per-node and per-attribute overrides take
precedence for the current session.

The node properties sidebar now includes a Visibility section with
controls for:
- hiding or showing the selected node
- toggling names
- toggling attributes
- toggling inputs
- toggling outputs
- toggling weights
- toggling connections
- showing all visibility criteria
- hiding all visibility criteria
- resetting visibility for the current node

The Attributes section in the node sidebar now includes an individual
checkbox beside each attribute row so users can control whether that
specific attribute is rendered. These controls are backed by per-
attribute session state and are reflected in graph redraws.

Inline graph controls were added to node headers so the user can open
node visibility controls quickly from the graph itself and hide a node
directly from the node chrome.

This change also adds a session-scoped display-name override layer.
Users can edit names directly in the sidebar wherever the UI has a
meaningful named target, including model names, target names, node
names, and sidebar field labels such as attributes, inputs, outputs,
and blocks. These edits are applied to the active session only and do
not modify the source model file.

UI support was added in source/index.html for the new visibility
buttons, per-attribute checkboxes, and editable text inputs, while
preserving the existing Netron visual style in both light and dark
themes.

Key implementation details:
- add VisibilityState, NodeVisibility, and FieldVisibility helpers
- track per-node and per-attribute UI visibility outside parsed model data
- reload and rebuild the graph after visibility or rename changes
- keep edits session-only with no persistence or file format changes
- preserve existing global view toggles as defaults

Validation performed:
- syntax check passed with `node --check source/view.js`

Known limitations:
- edits are session-only and reset when the model is reopened
- runtime behavior has not yet been fully smoke-tested in the browser or
  Electron app with multiple real models

### DIFF
--- a/source/index.html
+++ b/source/index.html
@@ -226,6 +226,7 @@ button { font-family: -apple-system, BlinkMacSystemFont, "Segoe WPC", "Segoe UI"
 .sidebar-item-value-line-link:hover { text-decoration: underline; }
 .sidebar-item-value-line-border { padding: 4px 6px 4px 6px; border-top: 1px solid rgba(27, 31, 35, 0.05); }
 .sidebar-item-value-line-content { white-space: pre; word-wrap: normal; overflow: auto; display: block; }
+.sidebar-editable-input { width: 100%; border: 0; outline: none; background: transparent; color: inherit; font: inherit; padding: 0; }
 .sidebar-item-value-expander { font-family: 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace; float: right; color: #aaa; cursor: pointer; user-select: none; -webkit-user-select: none; -moz-user-select: none; padding: 4px 6px 4px 4px; }
 .sidebar-item-value-expander:hover { color: #333; }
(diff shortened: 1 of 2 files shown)


spsrathor:
It can be used to:

- hide/show nodes
- toggle names, attributes, inputs, outputs, weights, and connections per node
- hide/show individual attributes
- temporarily rename display labels in the sidebar and graph UI for easier inspection

This is useful for exploring large or noisy models, reducing clutter during debugging, and creating clearer temporary views for analysis or demos. Changes are applied by rebuilding the rendered graph from UI state, not by mutating parsed model data, and reset when the session ends.

PR #1580, opened 2026-06-13
Fixed ManifestVersion in package.js.

DandelionSprout:
The newest ManifestVersion is 1.12.0, so I upgraded the script to use that version.

### DIFF
--- a/package.js
+++ b/package.js
@@ -414,16 +414,16 @@ const publish = async (target) => {
             writeLine('update manifest');
             const manifestFile = dirname(...paths, identifier);
             await fs.writeFile(`${manifestFile}.yaml`, [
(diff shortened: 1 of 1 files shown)


lutzroeder:
@DandelionSprout thank you. Is there a specific reason this is needed?

DandelionSprout:
There's been some worries among Winget-pkgs contributors recently about an increase in PRs that use 1.6.0 instead of anything else (https://github.com/microsoft/winget-pkgs/issues/380264).

Although the worry is specifically about new packages, in contrast to Netron which has been in the repo for some time already, there are no particular reasons to *not* change it to 1.12.0, and as far as I can tell there have been no breaking changes between those versions when it comes to x86 and ARM64.

(There are a handful of extremely rare fringe scenarios where 1.10.0 would make more sense, but Netron is not one of them.)

PR #1583, opened 2026-07-07
add docker deployment

LuminWolf:


lutzroeder:
Thank you @LuminWolf. Better built on the `pip install netron` package downstream than owned here.

Issue #697, opened 2021-03-05
Allow loading models from file url


JarnoRFB:
Very cool app!

Momentarily it is possible to load model from a http url like https://netron.app/?url=https://raw.githubusercontent.com/srom/chessbot/master/model/chessbot.pb 

It would be very useful if the model file could also be specified via a local file url like  https://netron.app/?url=file:/home/myuser/chessbot.pb 

The usecase I have is that I want to track model architectures in mlflow. So currently I can easily save my architecture in the mlflow ftp artifact store. Next I would like to run a private netron instance that has access to that artifact store mounted on its file system and save a url together with the model architecture in mlflow, so I can open the architecture with one click.

Also happy to implement this if the feature sounds useful, I would probably need a few pointers where to look though.

lutzroeder:
@JarnoRFB can you clarify how this is expected to work and maybe submit a pull request? Since the app is running in the browser which is a sandbox without access to the file system, how is this expected to work?

JarnoRFB:
@lutzroeder  Not sure if I understand you correctly. I mean one does load models from the file system via the UI or one can specify the path to the model file from the command line when starting the server. Therefore I thought that it would be possible to utilize one of those mechanisms to also load models from the file system specified via a URL. 

If you think that this is possible I can attempt an implementation.

JarnoRFB:
@lutzroeder Just noticed that mlflow also exposes its artifacts via its http API, so an access via a local file URL is no longer required for my usecase.

lutzroeder:
> Not sure if I understand you correctly. I mean one does load models from the file system via the UI or one can specify the path to the model file from the command line when starting the server.

First approach is the browser exposing a dialog or drag & drop gesture which requires user interaction. Second approach requires running a server. Neither allows the code running in the browser to directly access the local file system.

Issue #1007, opened 2022-11-22
Passing a model to browser app via data URL in URL fragment


FranklinWhale:
I would like to share a number of model to students and let them visualize it with Netron. Although there is a `url` parameter, it will be sent to the hosting server and the length may be limited. I hope the model can be passed to the browser application via URL fragment.

Example:
[...]

PR #1591, opened 2026-07-29
Add a postMessage embedding protocol to the browser host

take-cheeze:
## Summary

Rel https://github.com/lutzroeder/netron/issues/697 https://github.com/onnxsim/onnxsim/pull/519
Needed in model preview on https://onnxsim.github.io/onnxsim/ .

Lets an embedder hand a model to the web build of Netron **directly as bytes**
over `window.postMessage`, without uploading it or packing it into the URL.
This is modeled on
[Perfetto's embedding protocol](https://perfetto.dev/docs/visualization/embedding-the-ui).

The motivation is the browser URL-length limit: Chromium refuses to navigate to
URLs longer than ~2 MB (`about:blank#blocked`), so the existing
`#<data-url>` loading path fails for any model over ~1.5 MB. Passing the bytes
through a message channel removes that ceiling and keeps everything in-process —
nothing is uploaded.

## Protocol

1. The embedder opens Netron in a popup or `<iframe>`, ideally with the `?embed`
   flag (see below), and posts the string `"PING"` on an interval.
2. Netron replies `"PONG"` once its view is ready to receive a model.
3. The embedder posts `{ netron: { buffer, name, identifier, url } }`, where
   `buffer` is an `ArrayBuffer`/typed array of the raw model bytes.
4. Netron opens it and posts back `{ netron: { status: "success", identifier } }`
   or `{ netron: { status: "error", message } }`.

The bytes are loaded through the same in-memory path `_openGist` already uses
(`BinaryStream` → `Context` → `_openContext`).

## Changes

- **`source/browser.js`**
  - Install a `message` listener in the constructor, but only answer once the
    view is initialized (`start()`), so the handshake can never hand a model to
    a half-initialized view. A model that arrives early is stashed and opened
    when `start()` runs.
  - Reply `"PONG"` to `"PING"`; accept a `{ netron: { … } }` request; ignore all
    other `postMessage` traffic (extensions, dev tooling) by shape.
  - **Embedded mode** (`?embed`): suppress the first-run update/cookie-consent
    dialogs and skip telemetry startup, so an embedder can drive the window with
    no dialog in front of the model and no analytics firing without a user
    present. Skipping telemetry leaves the session unstarted, so `send()` no-ops
    on its own.
- **`docs/embedding.md`** — protocol spec with popup and iframe examples.
- **`docs/embedding-example.html`** — a runnable manual-test page (pick a file →
  embed Netron → handshake → load), with a configurable origin field.

## Notes

- Netron only *renders* a model (never executes it), so the request is accepted
  from any origin; `"PONG"` is sent back to `event.source` at the ping's own
  `event.origin`.
- No behavior change to the normal URL/`gist`/drag-drop loading paths, or to any
  non-web (Electron/Python) host.

## Testing

- `npm run lint` clean.
- `node package.js build web` builds; embedded mode present in the output.
- Manual: served the `web` build, drove it via `docs/embedding-example.html`
  (iframe + popup), verified the PING/PONG handshake and model load in embedded
  mode.

### DIFF
--- a/docs/embedding-example.html
+++ b/docs/embedding-example.html
@@ -1,0 +1,73 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Netron embedding example</title>
+    <style>
+        body { font-family: sans-serif; margin: 1em; }
+        #row { display: flex; gap: 0.5em; align-items: center; flex-wrap: wrap; margin-bottom: 0.5em; }
+        #origin { width: 22em; }
+        #status { color: #555; margin-bottom: 0.5em; min-height: 1.2em; }
+        iframe { width: 100%; height: 80vh; border: 1px solid #ccc; }
+    </style>
+</head>
+<body>
+    <!--
+      Manual test for the postMessage embedding protocol (see embedding.md).
+      Pick a model file; it is read in-page and handed to the embedded Netron as
+      raw bytes over postMessage — nothing is uploaded. Point "Netron origin" at
+      a build that includes the protocol (e.g. this fork's preview deployment);
+      upstream netron.app will not respond to the handshake until it ships.
+    -->
+    <div id="row">
+        <label>Netron origin: <input id="origin" value="https://netron.app"></label>
+        <input id="file" type="file">
+    </div>
+    <div id="status">Pick a model file to load it into the embedded Netron.</div>
+    <iframe id="frame" title="Netron"></iframe>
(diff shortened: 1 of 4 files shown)


lutzroeder:
Thank you. Netron is an app, not a component. This would add a protocol to keep stable and secure, which isn't a supported scenario. #1007 was intended for sharing small models with students. For large models providing a CORS location is supported.

